PR #11, opened 2017-01-31
celery: fix broker service name

diegodelemos:

### DIFF
--- a/reana_workflow_controller/tasks.py
+++ b/reana_workflow_controller/tasks.py
@@ -30,7 +30,8 @@ from celery import Celery
 
 
 celery = Celery('tasks',
-                broker='amqp://test:1234@message-broker//')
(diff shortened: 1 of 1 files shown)


reana_workflow_controller/tasks.py
```diff
@@ -30,7 +30,8 @@
 
 
 celery = Celery('tasks',
-                broker='amqp://test:1234@message-broker//')
+                broker='amqp://test:1234@'
+                       'message-broker.default.svc.cluster.local//')
```
tiborsimko: I guess we should amend this also in `reana-resources-k8s` in `deployments/reana-system/flower.yaml`?

diegodelemos: Indeed, I am preparing the PR.

PR #19, opened 2017-02-15
docker: fix Dockerfile command for executing flask

diegodelemos:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,4 +26,5 @@ RUN adduser --uid 1000 --disabled-password --gecos '' reanauser && \
     chown -R reanauser:reanauser /code
 USER reanauser
 EXPOSE 5000
-CMD ["python", "reana_workflow_controller/app.py"]
+ENV FLASK_APP reana_workflow_controller/app.py
+CMD ["flask", "run"]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -26,4 +26,6 @@ RUN adduser --uid 1000 --disabled-password --gecos '' reanauser && \
     chown -R reanauser:reanauser /code
 USER reanauser
 EXPOSE 5000
-CMD ["python", "reana_workflow_controller/app.py"]
+ENV FLASK_APP reana_workflow_controller/app.py
+ENV FLASK_DEBUG 1
```
tiborsimko: Do we want `FLASK_DEBUG` to be on by default? I think it's better to leave it out, and let the user set it on if wanted...

setup.py
```diff
@@ -64,7 +64,7 @@
 
 install_requires = [
     'celery==3.1.17',
-    'Flask==0.10.1',
+    'Flask>=0.12',
```
tiborsimko: Also, now that we require Flask 0.12, the commit is no longer about docker only, so it would be better to change to the commit log message to something like `installation: upgrade to Flask 0.12`.

PR #21, opened 2017-03-24
api: Add MOCKED /workflows -endpoint

### LINKED ISSUE #20: api: add workflows endpoint
Right now there is no quick way of checking whether `reana-workflow-controller` instance is working. We should add a `workflows` endpoint which should return an empty list of workflows for now such as `reana-job-controller` `jobs` endpoint.

hjhsalo:
* Add MOCKED /workflows -endpoint that provides a quick way to
  check if reana-workflow-engine is running. Closes #20.

Signed-off-by: Harri Hirvonsalo <harri.hirvonsalo@cern.ch>

### DIFF
--- a/reana_workflow_controller/app.py
+++ b/reana_workflow_controller/app.py
@@ -42,6 +42,46 @@ experiment_to_queue = {
 }
 
 
+@app.route('/workflows', methods=['GET'])
+def get_workflows():
+    """Get all workflows.
+
+    .. http:get:: /workflows
+
+        Returns a JSON list with all the workflows.
+
+        .. warning::
+            Not implemented, just a mock that can be used e.g. to quickly
+            check if reana-workflow-controller is running.
+
+        **Request**:
+
+        .. sourcecode:: http
+
+            GET /workflows HTTP/1.1
+            Content-Type: application/json
+            Host: localhost:5000
+
+        :reqheader Content-Type: application/json
+
+        **Responses**:
+
+        .. sourcecode:: http
+
+            HTTP/1.1 200 OK
+            Content-Length: 22
+            Content-Type: application/json
+
+            {
+              "workflows": {}
+            }
+
+        :resheader Content-Type: application/json
+        :statuscode 200: no error - the list has been returned.
+    """
+    return jsonify({"workflows": {}}), 200
+
+
 @app.route('/yadage', methods=['POST'])
 def yadage_endpoint():
     """Create a new job.
--- a/docs/restapi.rst
+++ b/docs/restapi.rst
@@ -1,7 +1,15 @@
 REST API
 ========
 
+GET /workflows
+--------------
+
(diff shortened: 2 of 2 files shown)


reana_workflow_controller/app.py
```diff
@@ -42,6 +42,34 @@
 }
 
 
+@app.route('/workflows', methods=['GET'])
+def get_workflows():
+    """Get all workflows.
+
```
diegodelemos: PEP8 expects here and empty line. Try to plug some linter to your editor so you notice this kind of errors before! :)

reana_workflow_controller/app.py
```diff
@@ -42,6 +42,34 @@
 }
 
 
+@app.route('/workflows', methods=['GET'])
+def get_workflows():
+    """Get all workflows.
+    
+    .. http:get:: /workflows
+        Returns a JSON list with all the workflows.
```
diegodelemos: I would suggest to keep empty lines between different blocks when documenting as in `yadage_endpoint()` docstring, in favor of readability

reana_workflow_controller/app.py
```diff
@@ -42,6 +42,34 @@
...
+        NOTE: Currently just a mock that can used e.g. to quickly
```
diegodelemos: Regarding to the NOTE, it is rendered all in the same line and it doesn't look very pretty. I would do either one of the following things:
- Add empty line between the endpoint description and the note.
```
        Returns a JSON list with all the workflows.

        NOTE: Currently just a mock that can used e.g. to quickly
        check if reana-workflow-controller is running.
```
- Use `sphinx` `.. warning::` so it gets nicely rendered with red background.
```
        .. warning::
            Not implemented, just a mock that can be used e.g. to quickly
            check if reana-workflow-controller is running.
```
p.s. mind the missing verb in `... just a mock that can [be] used ...`

reana_workflow_controller/app.py
```diff
@@ -42,6 +42,34 @@
...
+        .. sourcecode:: http
+            HTTP/1.1 200 OK
+            Content-Length: 22
+            Content-Type: application/json
+            {
+              "workflows": {}
+            }
+        :resheader Content-Type: application/json
+        :statuscode 200: no error - the list has been returned.
+    """
```
diegodelemos: Remember that you should reference this new endpoint in https://raw.githubusercontent.com/reanahub/reana-workflow-controller/master/docs/restapi.rst otherwise it doesn't get rendered

tiborsimko:
@hjhsalo Please squash the commits together at the end and use something like the following commit message style:

```
 api: mocked `/workflows` endpoint

* Adds mocked `/workflows` endpoint that provides a quick way to
  check if reana-workflow-engine is running. (closes #20)
``` 

This could be important later for automatic generation of release notes...

PR #26, opened 2017-06-07
persistence: add tenant db

diegodelemos:


reana_workflow_controller/api.py
```diff
@@ -0,0 +1,202 @@
...
+#
+# You should have received a copy of the GNU General Public License along with
+# REANA; if not, write to the Free Software Foundation, Inc., 59 Temple Place,
+# Suite 330, Boston, MA 02111-1307, USA.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization or
+# submit itself to any jurisdiction.
+
+"""REANA Workflow Controller Rest API."""
```
tiborsimko: It may be good to ponder renaming of `api.py` to `restapi.py`, so that we distinguish more clearly here and in other REANA packages between:

- API = Pythonic API, CLI API
- RESTAPI = REST API

While some REANA components won't have much of a Pythonic API, and they will be used exclusively over REST API, it may be good to have some naming consistency across all packages for ease of development across packages?

TBD @diegodelemos @hjhsalo

reana_workflow_controller/app.py
```diff
@@ -24,174 +24,28 @@
...
 
-experiment_to_queue = {
-    'alice': 'alice-queue',
-    'atlas': 'atlas-queue',
-    'lhcb': 'lhcb-queue',
-    'cms': 'cms-queue',
-    'recast': 'recast-queue'
-}
 
+def create_app():
```
tiborsimko: As discussed IRL, let's put factory creation into `factory.py` and let's leave `app.py` to instantiate the application out of it.

reana_workflow_controller/models.py
```diff
@@ -0,0 +1,55 @@
...
+
+from sqlalchemy.orm import validates
+
+from .app import db
+
+
+class Tenant(db.Model):
+    """Tenant model."""
+
+    id_ = db.Column(db.String(36), primary_key=True)
```
tiborsimko: `sqlalchemy-utils` has a `UUIDType` that we could use, see `invenio-records`.

reana_workflow_controller/models.py
```diff
@@ -0,0 +1,55 @@
...
+    email = db.Column(db.String(120), unique=True)
```
tiborsimko: Is 120 enough? Perhaps max out to 255?

reana_workflow_controller/models.py
```diff
@@ -0,0 +1,55 @@
...
+from __future__ import absolute_import
+
+import uuid
+
+from sqlalchemy.orm import validates
+
+from .app import db
+
+
+class Tenant(db.Model):
```
tiborsimko: We may want to have some user accounting information, so let's think of new columns like:

- create_date
- last_active_date

CC @diegodelemos @hjhsalo

tiborsimko: (and we could even think of future Alembic DB upgrade recipes 😄 to be fully future-proof)

tests/conftest.py
```diff
@@ -24,44 +24,71 @@
...
+
+    db.initialize_dbs()
+    yield db
+
+
+@pytest.yield_fixture()
+def default_tenant(db):
+    """Create users."""
+    default_tenant = Tenant('00000000-0000-0000-0000-000000000000',
+                            'default@reana.io',
```
tiborsimko: `info@reana.io`

tiborsimko:
(let's ignore coveralls since this code will go away soon, see #28)

PR #35, opened 2017-09-13
Get default user if already exists

anton-khodak:


diegodelemos:
could you please amend a bit the commit message? something like
`cli: detect if default user already exists` (before the `:` we try to write the component we are changing inside the package and after a short description) so if we have to search for it later we can easily find by keywords `cli`, `default user` or `already exists` thank you! 😃

anton-khodak:
Done!

anton-khodak:
yes, "detect" sounds clearer

diegodelemos:
`reanahub/reana-workflow-controller:v0.0.1` updated on DockerHub, thanks @anton-khodak!

PR #37, opened 2017-10-06
api: add workflow seeding endpoint

diegodelemos:

### DIFF
--- a/reana_workflow_controller/rest.py
+++ b/reana_workflow_controller/rest.py
@@ -25,6 +25,7 @@
 import traceback
 
 from flask import Blueprint, abort, jsonify, request
+from werkzeug.utils import secure_filename
 
 from .factory import db
 from .fsdb import get_all_workflows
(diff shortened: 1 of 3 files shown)


diegodelemos:
Remember to address https://github.com/reanahub/reana-workflow-engine-yadage/issues/32 since to be able to seed, the workspace needs to be created

reana_workflow_controller/rest.py
```diff
@@ -147,6 +148,79 @@ def get_workflows():  # noqa
...
+                "message": "Workflow has been added to the workspace",
+                "workflow_id": "cdcf48b1-c2f3-4693-8230-b066e088c6ac"
+              }
+        400:
+          description: >-
+            Request failed. The incoming data specification seems malformed
+    """
+    try:
+        file_ = request.files['file_content'].stream.read()
+        file_name = secure_filename(request.args['file_name'])
```
hjhsalo: Adding this here just as  a reminder:
> The function might return an empty filename. It’s your responsibility to ensure that the filename is unique and that you generate random filename if the function returned an empty one.
(http://werkzeug.pocoo.org/docs/0.12/utils/#werkzeug.utils.secure_filename)

diegodelemos: thanks 😄 , linked to the workspace creation issue (when we will write files to disk)

PR #40, opened 2017-10-20
api: add initialize workflow endpoint

diegodelemos:
* (Closes reanahub/reana-workflow-engine-yadage#32)

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -103,6 +103,86 @@
           }
         },
         "summary": "Returns all workflows."
+      },
+      "post": {
+        "description": "This resource expects a POST call to create a new workflow workspace.",
+        "operationId": "create_workflow",
+        "parameters": [
+          {
+            "description": "Required. Organization which the worklow belongs to.",
+            "in": "query",
+            "name": "organization",
+            "required": true,
+            "type": "string"
+          },
+          {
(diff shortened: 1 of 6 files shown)


docs/openapi.json
```diff
@@ -157,8 +206,7 @@
             "description": "Request succeeded. The file has been added to the workspace.",
             "examples": {
               "application/json": {
-                "message": "Workflow has been added to the workspace",
-                "workflow_id": "cdcf48b1-c2f3-4693-8230-b066e088c6ac"
+                "message": "The file input.txt has been successfully transferred."
```
tiborsimko: Better `input.csv` or `input.dat` instead of `.txt`... but we can use some consistent examples across REANA, see `mydata.csv`, `mycode.py` etc used in musings elsewhere.

reana_workflow_controller/rest.py
```diff
@@ -169,6 +172,105 @@ def get_workflows():  # noqa
...
+            properties:
+              message:
+                type: string
+              workflow_id:
+                type: string
+          examples:
+            application/json:
+              {
+                "message": "Workflow workspace has been created.",
+                "workflow_id": "cdcf48b1-c2f3-4693-8230-b066e088c6ac"
```
hjhsalo: Will this payload include a name of the workflow in the future?

PR #50, opened 2017-11-08
api: add set_workflow_status function

anton-khodak:


reana_workflow_controller/rest.py
```diff
@@ -918,7 +921,125 @@ def get_workflow_status(workflow_id):  # noqa
...
+        500:
+          description: >-
+            Request failed. Internal controller error.
+    """
 
+    try:
+        organization = request.args['organization']
+        user_uuid = request.args['user']
+        workflow = Workflow.query.filter(Workflow.id_ == workflow_id).first()
+        status = request.json
```
diegodelemos: We could check here if the status sent is correct and send a 400 if it is not. You have three variables representing status on the top of the file, you could make a list with the values and check the incoming string belongs to it, what do you think?

Also, did you explore the possibility of adding new tests to check this scenarios?

reana_workflow_controller/rest.py
```diff
@@ -918,7 +921,125 @@ def get_workflow_status(workflow_id):  # noqa
...
+            return jsonify({'message': 'Workflow {} does not exist'.
+                            format(workflow_id)}), 404
+        if not str(workflow.owner_id) == user_uuid:
+            return jsonify(
+                {'message': 'User {} is not allowed to access workflow {}'
+                 .format(user_uuid, workflow_id)}), 403
+        if status == 'start':
+            return start_workflow(organization, user_uuid, workflow)
         return jsonify({'id': workflow.id_,
                         'status': workflow.status.name,
```
diegodelemos: This status will be the previous (the one you just got from DB), one would expect to retrieve the requested status. I would return, for now, the requested status as it is (as long as it is correct). 

I know it is misleading (if you want to start, it could perfectly happen that it fails) but since we are calling a celery task which doesn't return any value and until [this](https://github.com/reanahub/reana-workflow-engine-yadage/blob/master/reana_workflow_engine_yadage/tasks.py#L109) gets executed in the workflow engine there won't be any change in DB we can't do anything else but returning the value as it is.

reana_workflow_controller/rest.py
anton-khodak: I agree, enumeration checks and tests are necessary

reana_workflow_controller/rest.py
anton-khodak: I will return the requested status for now, but we could extend the set of statuses and return here "starting" or "pending", for example, and "started" when the workflow starts in the workflow engine.

reana_workflow_controller/rest.py
```diff
@@ -927,3 +934,178 @@ def get_workflow_status(workflow_id):  # noqa
...
+                    'user': user_uuid}), 200
+
+
+def run_yadage_workflow_from_spec(organization, user_uuid, workflow):
+    """Run a yadage workflow."""
+    try:
+        kwargs = {
+            "workflow_workspace": workflow.workspace_path,
+            "workflow_json": workflow.specification,
+            "parameters": workflow.parameters
```
diegodelemos: `workflow_uuid` is missing see [here](https://github.com/reanahub/reana-workflow-engine-yadage/blob/master/reana_workflow_engine_yadage/tasks.py#L69). This will probably break when calling celery, have you tried to do the following with all your PRs integrated:
```console
~/reana-demo-root6-roofit $ reana-client -l debug workflow create -f .reana.yaml
some-uuid
~/reana-demo-root6-roofit $ reana-client workflow start --workflow some-uuid
workflow started blah blah
```
And check if it actually works?

anton-khodak: Indeed, I just copied the code from `run_yadage_workflow_from_spec_endpoint`, it is fixed. However, I have troubles with spawning yadage workers both without kubernetes and inside minikube, so I couldn't test the full chain. 

(the trouble is in connecting to message broker, I have UnicodeDecodeError, would be good if you could verify running/failing engine-yadage later. I don't have such a problem with engine-cwl)

reana_workflow_controller/rest.py
```diff
@@ -43,6 +48,10 @@
     'default': 'default-queue'
 }
 
+workflow_spec_to_task = {
```
diegodelemos: unused?

anton-khodak: leftovers, thanks

reana_workflow_controller/rest.py
```diff
@@ -927,3 +934,179 @@ def get_workflow_status(workflow_id):  # noqa
...
+        if not workflow:
+            return jsonify({'message': 'Workflow {} does not exist'.
+                            format(workflow_id)}), 404
+        if not str(workflow.owner_id) == user_uuid:
+            return jsonify(
+                {'message': 'User {} is not allowed to access workflow {}'
+                 .format(user_uuid, workflow_id)}), 403
+        if status == START:
+            return start_workflow(organization, user_uuid, workflow)
+        elif status == STOP:
```
diegodelemos: This is not implemented yet. I would remove it since changing the status in DB won't do anything, if the workflow is running it will keep running anyways and user will see `finished` status while it is not true.

reana_workflow_controller/rest.py
```diff
@@ -927,3 +934,179 @@ def get_workflow_status(workflow_id):  # noqa
...
+    db.session.commit()
+    if workflow.type_ == 'yadage':
+        return run_yadage_workflow_from_spec(organization,
+                                             user_uuid,
+                                             workflow)
+    elif workflow.type_ == 'cwl':
+        pass
+
+
+def stop_workflow(organization, user_uuid, workflow):
```
diegodelemos: Not implemented, and probably it won't be the way of stopping it (through DB)

reana_workflow_controller/rest.py
anton-khodak: I created it to test status changes similar to `get_workflow_status` tests. If `stop` is not on your roadmap in the nearest future, I will remove it.

diegodelemos: Indeed it is in the roadmap, but still not implemented so this "changing in the DB directly" would be kind of a "lie" because it is not how it will be implemented most probably.

If you want to check status changes you can get the status right after creating the workflow, which should be `created`, and then submit status change for `started` so you check if it works.

diegodelemos:
once amended those small changes, could you please squash the commits since everything belongs to the same functionality added? then we :shipit:

anton-khodak:
I couldn't figure out how to squash the commits with a prior merge commit, so I am switching to a new clean branch #51

PR #53, opened 2017-11-24
Reana CWL runner

anton-khodak:


run-tests.sh
```diff
@@ -21,7 +21,6 @@
 # submit itself to any jurisdiction.
 
 pydocstyle reana_workflow_controller && \
-isort -rc -c -df **/*.py && \
```
jirikuncar: ❓

anton-khodak: I have weird issues with isorts so I deleted them out for my testing purposes but forgot not to include them into commits. Will rectify that

PR #55, opened 2018-01-12
api: return 404 when file does not exist

diegodelemos:


docs/openapi.json
```diff
@@ -652,10 +652,10 @@
             "description": "Request failed. The incoming data specification seems malformed."
           },
           "404": {
-            "description": "Request failed. User doesn't exist.",
+            "description": "Request failed. `file_name` does not exist.",
             "examples": {
```
tiborsimko: Please amend also surrounding inputs/outputs cases...

PR #59, opened 2018-01-16
api: fix returning 500 when workflow doesn't exist

diegodelemos:


diegodelemos:
@tiborsimko tests passing locally

PR #61, opened 2018-01-18
api: refactor listing/seeding files endpoints

diegodelemos:


reana_workflow_controller/rest.py
```diff
@@ -521,103 +519,19 @@ def get_workflow_inputs(workflow_id):  # noqa
...
-      produces:
-        - multipart/form-data
-      parameters:
-        - name: organization
-          in: query
-          description: Required. Organization which the worklow belongs to.
-          required: true
-          type: string
-        - name: user
+        - name: input_type
```
hjhsalo: The `get_workflow_files` endpoint is used to retrieve files (either type `code`, `input`, `output`) of a workflow / analysis.
The parameter `input_type` may not be the most descriptive term to specify that one wants to retrieve all files that have been created by a workflow.
Maybe `file_type` or just `type` would be more descriptive or not less confusing compared to `input_type`, but on the other hand `file_type` or `type` is kind of ambiguous.

Parameter `input_type` in `seed_workflow_workspace` endpoint doesn't need to be renamed since it's very descriptive in that context, but on the other hand, we could promote consistency and rename it to `file_type` or `type` as well.

Sorry I don't have better suggestions regarding to renaming the parameter...

PR #68, opened 2018-02-27
api: detect when workflow can not be started

diegodelemos:
* (closes reanahub/reana-client#75).

### DIFF
--- a/reana_workflow_controller/errors.py
+++ b/reana_workflow_controller/errors.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2018 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it under the
+# terms of the GNU General Public License as published by the Free Software
+# Foundation; either version 2 of the License, or (at your option) any later
+# version.
+#
+# REANA is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+# A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# REANA; if not, write to the Free Software Foundation, Inc., 59 Temple Place,
+# Suite 330, Boston, MA 02111-1307, USA.
+#
+# In applying this license, CERN does not waive the privileges and immunities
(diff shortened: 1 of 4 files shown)


reana_workflow_controller/errors.py
```diff
@@ -0,0 +1,26 @@
...
+# REANA; if not, write to the Free Software Foundation, Inc., 59 Temple Place,
+# Suite 330, Boston, MA 02111-1307, USA.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization or
+# submit itself to any jurisdiction.
+
+"""REANA Workflow Controller errors."""
+
+class UnableToStartWorkflowError(Exception):
```
tiborsimko: What about having only two errors, a "continuable error" and a "fatal error", and the concrete error message will be passed as argument?  (Trying to foresee a future so that the number of error messages such as "cannot start", "cannot get status", "cannot suspend", "cannot resume", etc wouldn't proliferate too much...)

reana_workflow_controller/rest.py
```diff
@@ -1195,24 +1215,39 @@ def set_workflow_status(workflow_id):  # noqa
...
+                                                 workflow)
+        elif workflow.type_ == 'cwl':
+            return run_cwl_workflow_from_spec_endpoint(organization,
+                                                       workflow)
+        else:
+            raise NotImplementedError(
+                'Workflow type {} is not supported.'.format(workflow.type_))
+    else:
+        verb = "is" if workflow.status == WorkflowStatus.running else "has"
+        message = \
```
hjhsalo: I think the message is fine as is (_"The response body SHOULD include enough information for the user to recognize the source of the conflict."_, [RFC2616 section 10.4.10 409 Conflict](https://tools.ietf.org/html/rfc2616#section-10.4.100)), but just want to check that is there anything more to add to the message? (_"Ideally, the response entity would include enough information for the user or user agent to fix the problem; however, that might not be possible and is not required."_, [RFC2616 section 10.4.10 409 Conflict](https://tools.ietf.org/html/rfc2616#section-10.4.100))

Does it need to say that stop the workflow or wait for it to finish or fail?
Or something like "Only workflows that are not running can be started"?

diegodelemos: I don't think that now we have more trustable information to give right now, once we have more available status changes we can extend it. If the rest is fine and it has been tested, it should be ready to 🚀

Issue #72, opened 2018-03-30
Run number increases between different workflow names


dinosk:
When one creates two workflows with different names the following is returned:
```
$ reana-client workflow create -n a
a.1
$ reana-client workflow create -n b
b.2
```
workflow `b` starts at run number 2 which is not right, especially since we specify that it depicts the run number:
```
reana-client workflow status --workflow b
name|run_number|id                                  |user                                |organization|status 
----|----------|------------------------------------|------------------------------------|------------|-------
b   |2         |901ff09b-6e15-4ebe-85cc-08fc32df4fad|00000000-0000-0000-0000-000000000000|default     |created
```

We can create a separate table for the workflow `Runs` to solve this, and split the information related to the workflow and the run results, or we can just manually increment the run number for rows with identical workflow names.

dinosk:
Also related is the following:
```
(reana-cwl) ➜  reana-demo-root6-roofit git:(master) ✗ reana-client workflow create
workflow.1
(reana-cwl) ➜  reana-demo-root6-roofit git:(master) ✗ reana-client workflow status --workflow workflow
name    |run_number|id                                  |user                                |organization|status 
--------|----------|------------------------------------|------------------------------------|------------|-------
workflow|1         |be1f834a-7e48-434a-a308-5ae8f8503bb7|00000000-0000-0000-0000-000000000000|default     |created

(reana-cwl) ➜  reana-demo-root6-roofit git:(master) ✗ reana-client workflow create -n foo           
foo.2
(reana-cwl) ➜  reana-demo-root6-roofit git:(master) ✗ reana-client workflow status --workflow foo.1 
name    |run_number|id                                  |user                                |organization|status 
--------|----------|------------------------------------|------------------------------------|------------|-------
workflow|1         |be1f834a-7e48-434a-a308-5ae8f8503bb7|00000000-0000-0000-0000-000000000000|default     |created
```

PR #74, opened 2018-04-10
api: fix querying for workflow status

dinosk:
* FIXES querying for workflow statuses when specifying
  the workflow name and the run number.

(Addresses https://github.com/reanahub/reana-workflow-controller/issues/72)

Signed-off-by: Dinos Kousidis <konstantinos.kousidis@cern.ch>

### DIFF
--- a/reana_workflow_controller/rest.py
+++ b/reana_workflow_controller/rest.py
@@ -414,6 +414,7 @@ def seed_workflow_workspace(workflow_id_or_name):
             Request failed. Internal controller error.
     """
     try:
+        user_uuid = request.args['user']
         file_ = request.files['file_content']
         # file_name = secure_filename(request.args['file_name'])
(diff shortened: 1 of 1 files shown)


reana_workflow_controller/rest.py
```diff
@@ -1509,7 +1509,9 @@ def _get_workflow_with_uuid_or_name(uuid_or_name):
         # Search by `run_number` since it is a primary key.
         # workflow = Workflow.query. \
         #     filter(Workflow.name == run_number).first()
-        workflow = Workflow.query.get(run_number)
+        workflow = Workflow.query.filter(Workflow.name == workflow_name,
+                                         Workflow.run_number == run_number).\
```
tiborsimko: We should probably explicitly include a test for `owner_id` in case several people use the same workflow name... especially in case we delete the auto-incremented run number...

PR #82, opened 2018-05-08
global: receive messages from broker

dinosk:
* ADDS background task to consume message queue for job statuses.

(Contains https://github.com/reanahub/reana-workflow-controller/pull/83, so we could close it if we merge this)

### DIFF
--- a/reana_workflow_controller/tasks.py
+++ b/reana_workflow_controller/tasks.py
@@ -30,7 +30,7 @@ import jsonpickle
 import pika
 from celery import Celery
 from reana_commons.database import Session
-from reana_commons.models import Workflow, WorkflowStatus
+from reana_commons.models import Run, Workflow, WorkflowStatus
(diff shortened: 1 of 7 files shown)


reana_workflow_controller/tasks.py
```diff
@@ -38,3 +45,53 @@
...
+            if 'message' in body_dict and \
+                    'current_step' in body_dict.get('message'):
+                Session.query(Run).filter_by(workflow_uuid=workflow_uuid).\
+                    update(dict(
+                        current_step=body_dict['message']['current_step'],
+                        current_command=body_dict['message']['cmd'],
+                        total_steps=body_dict['message']['total_steps']))
+            Session.commit()
+
+    # def _callback_job_progress(ch, method, properties, body):
```
dinosk: one sec, fixing this

PR #94, opened 2018-06-19
utils: explicitly import fs.path

diegodelemos:
* When `fs` module is imported it does not contain the path module,
  it has to be imported explicitly. This causes a rare case when,
  if in the scope the instruction `from fs import path` is executed
  then `path` is added to the `fs` object otherwise an AttributeError
  exception will be thrown (closes reanahub/reana-client#113).

### DIFF
--- a/reana_workflow_controller/utils.py
+++ b/reana_workflow_controller/utils.py
@@ -22,6 +22,7 @@
 """Workflow persistence management."""
 
 import fs
+import fs.path as fs_path
 from flask import current_app as app
 from reana_commons.utils import get_user_analyses_dir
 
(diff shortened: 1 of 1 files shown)


diegodelemos:
Test case:
```
 $ docker run -ti --rm reanahub/reana-workflow-controller:0.3.0.dev20180619 bash
root@991967f8dcdd:/code# python
Python 3.6.5 (default, Jun  6 2018, 19:19:24)
[GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import fs
>>> dir(fs)
['ResourceType', 'Seek', '__all__', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', '__version__', '_fscompat', '_version', 'enums', 'fsdecode', 'fsencode', 'open_fs', 'opener']
>>> fs.path
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'fs' has no attribute 'path'
>>> from fs import path
>>> dir(fs)
['ResourceType', 'Seek', '__all__', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', '__version__', '_fscompat', '_version', 'enums', 'errors', 'fsdecode', 'fsencode', 'open_fs', 'opener', 'path']
>>> fs.path
<module 'fs.path' from '/usr/local/lib/python3.6/site-packages/fs/path.py'>
```

PR #95, opened 2018-06-21
cli: add cli command to run consume-job-status

diegodelemos:


diegodelemos:
@tiborsimko @dinosk following IRL dicussion, tests pass here since we only use `reana-commons` for the models and we set [sqlite in memory](https://github.com/reanahub/reana-workflow-controller/blob/acbf03c9e69a523b6adb47e9efd99ddd1b852690/tests/conftest.py#L59) as configuration for the test app. Do we still want to switch to postgres for testing? I would leave sqlite per component and ese Postgres as default in commons (DB initialisation is going to end up in `reana-server` anyways).

tiborsimko:
@diegodelemos Sounds good for now. If/when we move init db away from commons, we can switch back to sqlite there as well... and reorganise things later during docker-based setup sprint.

diegodelemos:
ready @tiborsimko  @dinosk

PR #101, opened 2018-06-27
global: addition of MQ consumer class

dinosk:
* ADD consumer class with reconnection strategy based on
  pika usage example: https://pika.readthedocs.io/en/latest/examples/blocking_consume_recover_multiple_hosts.html

Addresses #102 

### DIFF
--- a/reana_workflow_controller/consumer.py
+++ b/reana_workflow_controller/consumer.py
@@ -1,0 +1,93 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2018 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it under the
+# terms of the GNU General Public License as published by the Free Software
+# Foundation; either version 2 of the License, or (at your option) any later
+# version.
+#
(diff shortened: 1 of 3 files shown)


reana_workflow_controller/consumer.py
```diff
@@ -0,0 +1,93 @@
...
+from reana_commons.database import Session
+from reana_commons.models import WorkflowStatus
+
+from .config import (BROKER_PASS, BROKER_PORT, BROKER_URL, BROKER_USER,
+                     STATUS_QUEUE)
+from .tasks import (_update_job_progress, _update_run_progress,
+                    _update_workflow_status)
+
+
+class Consumer:
```
diegodelemos: What do you think about renaming to `JobStatusConsumer` to better describe what it does?

tiborsimko: We could also think about container names `[rest job-status-consumer]` to be something like `[rest-api-server job-status-consumer]`.  I.e. both components being named "-er" in the same noun way. What about merging and amending all those names in another PR while @dinosk is away?

PR #104, opened 2018-07-01
global: receive and store job caching info

dinosk:
depends on https://github.com/reanahub/reana-commons/pull/22, https://github.com/reanahub/reana-cluster/pull/93, https://github.com/reanahub/reana-commons/pull/29
Closes https://github.com/reanahub/reana-job-controller/issues/76

### DIFF
--- a/reana_workflow_controller/tasks.py
+++ b/reana_workflow_controller/tasks.py
@@ -24,14 +24,12 @@
 
 from __future__ import absolute_import
 
-import json
 import uuid
 
-import hashlib
 from celery import Celery
 from reana_commons.database import Session
 from reana_commons.models import Job, JobCache, Run, RunJobs, Workflow
-from reana_commons.utils import calculate_hash_of_dir
+from reana_commons.utils import calculate_hash_of_dir, calculate_job_input_hash
 
 from reana_workflow_controller.config import BROKER, POSSIBLE_JOB_STATUSES
 
@@ -56,6 +54,9 @@ def _update_workflow_status(workflow_uuid, status, logs):
 def _update_run_progress(workflow_uuid, msg):
     """Register succeeded Jobs to DB."""
     run = Session.query(Run).filter_by(workflow_uuid=workflow_uuid).first()
+    cached_jobs = None
+    if "cached" in msg['progress']:
+        cached_jobs = msg['progress']['cached']
     for status in POSSIBLE_JOB_STATUSES:
         if status in msg['progress']:
             previous_total = getattr(run, status)
@@ -71,7 +72,9 @@ def _update_run_progress(workflow_uuid, msg):
                     job = Session.query(Job).\
                         filter_by(id_=job_id).one_or_none()
                     if job:
-                        if job.status != status:
+                        if job.status != status or \
+                                (cached_jobs and
+                                 str(job.id_) in cached_jobs['job_ids']):
                             new_total += 1
                 new_total += previous_total
                 setattr(run, status, new_total)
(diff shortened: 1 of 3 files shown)


reana_workflow_controller/tasks.py
```diff
@@ -99,3 +105,20 @@ def _update_job_progress(workflow_uuid, msg):
                     run_job.run_id = current_run.id_
                     run_job.job_id = job_id
                     Session.add(run_job)
+
+
+def _update_job_cache(msg):
+    """Update caching information for finished job."""
+    cmd = msg['caching_info']['job_spec']['cmd']
+    clean_cmd = cmd.split(';')[1]
```
dinosk: This is a "hack" to clean the commnands from `r-w-e-serial`, that have prepended: `cd <path-to-workspace>` which includes a unique id, making the command hash different each time.

diegodelemos: 👍  we can add a comment so we remember

reana_workflow_controller/tasks.py
```diff
@@ -99,3 +105,20 @@ def _update_job_progress(workflow_uuid, msg):
...
+    msg['caching_info']['job_spec']['cmd'] = clean_cmd
+    input_hash = calculate_job_input_hash(msg['caching_info']['job_spec'],
+                                          msg['caching_info']['workflow_json'])
+    cached_job = JobCache(
+        job_id=msg['caching_info'].get('job_id'),
+        parameters=input_hash,
+        result_path=msg['caching_info'].get('result_path'),
+        workspace_hash=calculate_hash_of_dir(
```
diegodelemos: As discussed IRL, we should not store the job cache if `workspace_hash` failed.

reana_workflow_controller/tasks.py
```diff
@@ -99,3 +105,24 @@ def _update_job_progress(workflow_uuid, msg):
...
+    """Update caching information for finished job."""
+    cmd = msg['caching_info']['job_spec']['cmd']
+    # removes cd to workspace, to be refactored
+    clean_cmd = cmd.split(';')[1]
+    msg['caching_info']['job_spec']['cmd'] = clean_cmd
+    input_hash = calculate_job_input_hash(msg['caching_info']['job_spec'],
+                                          msg['caching_info']['workflow_json'])
+    directory_hash = calculate_hash_of_dir(
+        msg['caching_info'].get('workflow_workspace').
+        replace('data', 'reana/default'))
```
diegodelemos: This is going to invalidate the cache for all non `default` organization jobs. Something like https://github.com/reanahub/reana-job-controller/pull/74/commits/a2a73e6bf01f708b3b6d27fc387df569c09d8a7d needs to be done. You can get the experiment/organization name from the `job_spec`, we have to be careful though since soon we will be changing `experiment` variable name to `organization`.

reana_workflow_controller/tasks.py
```diff
@@ -99,3 +105,24 @@ def _update_job_progress(workflow_uuid, msg):
...
+
+def _update_job_cache(msg):
+    """Update caching information for finished job."""
+    cmd = msg['caching_info']['job_spec']['cmd']
+    # removes cd to workspace, to be refactored
+    clean_cmd = cmd.split(';')[1]
+    msg['caching_info']['job_spec']['cmd'] = clean_cmd
+    input_hash = calculate_job_input_hash(msg['caching_info']['job_spec'],
+                                          msg['caching_info']['workflow_json'])
+    directory_hash = calculate_hash_of_dir(
```
diegodelemos: As spotted today, RJC controller calculates the hash before the job has been started (see [here](https://github.com/reanahub/reana-job-controller/pull/74/files#diff-8ee72e3f835364b2cb6e08b399497ce7R98)) but for steps that do thinks like `- echo 'Great it is Friday' > test.txt` which are fast to execute, the workflow workspace will change by the moment this [line](https://github.com/reanahub/reana-workflow-controller/pull/104/files#diff-d95dcaa8d671d2a3c798faa7fd757a3fR118) is executed, calculating a different `workspace_hash`

reana_workflow_controller/tasks.py
```diff
@@ -99,3 +107,42 @@ def _update_job_progress(workflow_uuid, msg):
...
+        for filename in cached_job.access_times:
+            if filename in file_access_times and \
+                    file_access_times[filename] > \
+                    cached_job.access_times[filename]:
+                input_files.append(filename)
+    else:
+        return
+    cmd = msg['caching_info']['job_spec']['cmd']
+    # removes cd to workspace, to be refactored
+    clean_cmd = cmd.split(';')[1]
```
diegodelemos: shall we `';'.join(test.split(';')[1:])` since most probably there will be more `;` inside the command?

dinosk: good point, fixed here and in `r-w-e-serial` 👍

Issue #103, opened 2018-06-29
class and container naming


tiborsimko:
Harmonise naming, e.g.class Consumer -> class JobStatusConsumer to be more specific and to align with container name `job-status-consumer`.

Amend also container name `rest` to something noun-based such as `rest-api-server` to conform to our usual noun-based naming convetions wotj components/workers.

(stemmed from https://github.com/reanahub/reana-workflow-controller/pull/101#discussion_r198765946)

PR #106, opened 2018-07-04
consumer: addition of JobStatusConsumer class

dinosk:
* ADD JobStatusConsumer class, which inherits from the
  base Consumer class in reana-workflow-commons and sets
  it to consumer from the jobs-status queue.

Depends on https://github.com/reanahub/reana-workflow-commons/pull/2
Addresses https://github.com/reanahub/reana-workflow-controller/issues/103


reana_workflow_controller/consumer.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2018 CERN.
+# Copyright (C) 2017, 2018 CERN.
```
diegodelemos: 2018 only?

reana_workflow_controller/consumer.py
```diff
@@ -20,39 +20,23 @@
 # granted to it by virtue of its status as an Intergovernmental Organization or
 # submit itself to any jurisdiction.
 
-"""REANA Workflow Controller MQ Consumer."""
```
diegodelemos: Missing module docstring.

Dockerfile
```diff
@@ -24,8 +24,6 @@ RUN apt-get update && \
     apt-get install -y vim-tiny && \
     pip install --upgrade pip
 
-RUN pip install -e git://github.com/reanahub/reana-commons.git@master#egg=reana-commons
```
diegodelemos: Missing `reana-commons`. You can add it just before `COPY . /code`, also `reana-workflow-commons` could be moved before.

dinosk: Adding it, in https://github.com/reanahub/reana-workflow-engine-cwl/pull/46 @tiborsimko suggested adding it after the `COPY . /code`, I am not sure if it affects the rebuilding though.

diegodelemos: Because until we don't release `reana-commons` and `reana-workflow-commons` in PyPI we will have to rebuild the image with no cache if we change any of the repos since the docker builder will take it as cached. You can move them after and once we release on PyPI we remove the lines, I just made the comment because it is pulling from Github the two repos and installing them every time I rebuild, therefore slowing down the rebuild.

Lets follow @tiborsimko advise then:
> I would propose the safest order like:
pypi requirements builder all to cache
COPY . /code
pip r-commons r-w-commons
pip install .

dinosk: I was referring to the order `1. COPY 2. pip install -e git://...` if it affected the docker image building, the rest is of course obvious. The order is already as described

Dockerfile
```diff
@@ -23,7 +23,7 @@ FROM python:3.6
 RUN apt-get update && \
     apt-get install -y vim-tiny && \
     pip install --upgrade pip
-
+RUN pip install -e git://github.com/reanahub/reana-workflow-commons.git@master#egg=reana-workflow-commons
```
dinosk: The 2 pip install from github had to be moved here, before the installation of requirements-builder, as `reana-commons` and `reana-workflow-commons` need to be preset in the `setup.py`, and it fails otherwise.

PR #113, opened 2018-07-23
rest: optional parameters field

dinosk:
Closes https://github.com/reanahub/reana-client/issues/109

### DIFF
--- a/reana_workflow_controller/rest.py
+++ b/reana_workflow_controller/rest.py
@@ -231,6 +231,7 @@ def create_workflow():  # noqa
               name:
                 type: string
                 description: Workflow name. If empty name will be generated.
+            required: [specification, type, name]
       responses:
         201:
           description: >-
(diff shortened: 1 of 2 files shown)


reana_workflow_controller/rest.py
```diff
@@ -291,7 +291,7 @@ def create_workflow():  # noqa
                             name=workflow_name,
                             owner_id=request.args['user'],
                             specification=request.json['specification'],
-                            parameters=request.json['parameters'],
+                            parameters=request.json.get('parameters'),
```
diegodelemos: this is a nice opportunity to update the docstrings using the `required` field as we use in [RJC](https://github.com/reanahub/reana-job-controller/blob/master/docs/openapi.json#L82) for optional properties.

dinosk: do you mean for the nested schema in https://github.com/dinosk/reana-workflow-controller/blob/27d46b78619a4692d51cbfc776261e8539659bc6/reana_workflow_controller/rest.py#L221?

diegodelemos: yes, in RJC it is easier since we are using marshmallow https://github.com/reanahub/reana-job-controller/blob/master/reana_job_controller/schemas.py, so this schemas are autogenerated. We should probably use it everywhere taking also advantage of validation, but for now lets do it manually.

PR #124, opened 2018-08-29
tasks: split workflow params from engine params

diegodelemos:
* Since we are interpolating user provided commands with their
  parameters (expanding any occurrence of $varname or ${varname}
  with the values provided as workflow parameters, we can not
  mix workflow params, those to be populated into the commands,
  and engine parameters, those to be passed to the engines to behave
  is supported and checked for correctness.
  differently (closes reanahub/reana-workflow-engine-serial#34).

### DIFF
--- a/reana_workflow_controller/rest.py
+++ b/reana_workflow_controller/rest.py
@@ -1357,14 +1357,15 @@ def run_cwl_workflow_from_spec_endpoint(workflow):  # noqa
         abort(400)
 
 
-def run_serial_workflow_from_spec(workflow, parameters):
+def run_serial_workflow_from_spec(workflow, engine_parameters):
     """Run a serial workflow."""
     try:
         kwargs = {
             "workflow_uuid": str(workflow.id_),
             "workflow_workspace": workflow.get_workspace(),
             "workflow_json": workflow.specification,
-            "parameters": {**workflow.parameters, **parameters},
+            "workflow_parameters": workflow.parameters,
+            "engine_parameters": engine_parameters,
         }
         if not os.environ.get("TESTS"):
             resultobject = run_serial_workflow.apply_async(


diegodelemos:
Related to https://github.com/reanahub/reana-client/pull/151#discussion_r208607561.

reana_workflow_controller/rest.py
```diff
@@ -1364,7 +1364,8 @@ def run_serial_workflow_from_spec(workflow, parameters):
             "workflow_uuid": str(workflow.id_),
             "workflow_workspace": workflow.get_workspace(),
             "workflow_json": workflow.specification,
-            "parameters": {**workflow.parameters, **parameters},
+            "parameters": workflow.parameters,
+            "engine_parameters": parameters,
```
tiborsimko: The new "engine parameters" seems to be akin to the what we have been calling in the past "operational parameters".  In other words, there are:

1. "workflow parameters" that affect the "science" of the workflow run (e.g. region name, year min, year max for our world population analysis example);

2. "operational parameters" that do not modify the science but specify how the workflow should be processed from a operational point of view  (e.g. run with/without cache, e.g. split into 10 or 100 parallel tasks, e.g. use big RAM, etc.) 

Perhaps we could reflect this optics in the API and `engine_parameters` could become `operational_parameters`?

lukasheinrich: iirc we discussed something similar in the context of e.g. backend options in yadage that map alias names for images to actual docker images. In general I think it's fair to assume that each engine has a wide range of options, which should be possible to be set via REANA yml. to me  `engine_options` or `engine_parameters` communicates the intent more clearly than `operational_parameters`

diegodelemos: > to me `engine_options` or `engine_parameters` communicates the intent more clearly than `operational_parameters`

I agree, I think that here we should keep `engine_parameters` as it is deep in the codebase and it is related to the next component which is always going to be a REANA Workflow **Engine** __ , so from a programmer point of view it makes sense. 

> Perhaps we could reflect this optics in the API and `engine_parameters` could become `operational_parameters`?

I agree as well, `operational_parameters` is more appropriate for REST API and REANA yaml since for end users an engine, name that we chose to call this entity, doesn't say that much. Whether we call it `operational_parameters` or something else, we have to be descriptive about what it is, mentioning engine would mean that people should be familiar with the architecture, which shouldn't be the case.

PR #125, opened 2018-08-29
 rest: harmonise time format

diegodelemos:
* Renames config variable since it no longer concerns workflows alone
  but also files (closes reanahub/reana-client#162).

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -107,37 +107,29 @@
               "application/json": [
                 {
                   "created": "2018-06-13T09:47:35.66097",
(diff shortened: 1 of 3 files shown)


diegodelemos:
Switching to ISO8601 for all REST calls see:
- https://swagger.io/specification/#data-types
- https://json-schema.org/understanding-json-schema/reference/string.html#built-in-formats
- https://xml2rfc.tools.ietf.org/public/rfc/html/rfc3339.html#anchor14
- https://www.w3.org/TR/NOTE-datetime

PR #133, opened 2018-10-04
consumer: use Kombu based consumer

### LINKED ISSUE #131: tests: mock AMQP to test consumer.py
Currently [consumer.py](https://github.com/reanahub/reana-workflow-controller/blob/2f8f7814951c6664785ff38eda238973d4ac2d9c/reana_workflow_controller/consumer.py) is not being tested, the right infrastructure should be put in place to test this behaviour.

diegodelemos:
Closes #131.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,6 @@ COPY . /code
 
 # Debug off by default
 ARG DEBUG=false
-
 RUN if [ "${DEBUG}" = "true" ]; then pip install -r requirements-dev.txt; pip install -e .; else pip install .; fi;
 
 EXPOSE 5000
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -26,6 +26,9 @@ RUN apt-get update && \
 
 COPY CHANGES.rst README.rst setup.py /code/
 COPY reana_workflow_controller/version.py /code/reana_workflow_controller/
+ARG nocache
+RUN echo $nocache
```
dinosk: Could $nocache be removed?

diegodelemos: absolutly! I was using it locally to not wait more on the rebuilds!

PR #141, opened 2018-11-01
api: addition of deleted status

dinosk:
* ADDS deleted status and handler for
  workflow deletion in set_workflow_status().

Depends on https://github.com/reanahub/reana-db/pull/15
Depends on https://github.com/reanahub/reana-commons/pull/67
Connects https://github.com/reanahub/reana-client/issues/192


reana_workflow_controller/rest.py
```diff
@@ -1526,3 +1528,57 @@ def _get_current_job_progress(workflow_id):
...
+def _hard_delete_workflow(workflow):
+    """Remove workflow row from database."""
+    remove_workflow_workspace(workflow.get_workspace())
+    Session.query(Workflow).filter_by(id_=workflow.id_).delete()
+    Session.commit()
+
+
+def _soft_delete_workflow(workflow):
+    """Mark workflow as deleted."""
+    remove_workflow_workspace(workflow.get_workspace())
```
dinosk: We could consider leaving the workspace in place for a short duration of time (like 1 day) and removing it afterwards, to allow users to download files from it.

reana_workflow_controller/rest.py
```diff
@@ -24,23 +24,21 @@
...
 START = 'start'
 STOP = 'stop'
 PAUSE = 'pause'
-STATUSES = {START, STOP, PAUSE}
-
-workflow_spec_to_task = {
-    "yadage": run_yadage_workflow,
-    "cwl": run_cwl_workflow
-}
+DELETED = 'deleted'
```
diegodelemos: Shall we improve this and use the [WorkflowStatus enum](https://github.com/reanahub/reana-db/blob/master/reana_db/models.py#L56-L62) instead of hardcoding this strings here? We could do `[s.name for s in WorkflowStatus]`.

dinosk: Yes I agree, since this was not introduced in this PR I'll create a new issue.

reana_workflow_controller/rest.py
```diff
@@ -1226,11 +1224,15 @@ def set_workflow_status(workflow_id_or_name):  # noqa
...
                  .format(user_uuid, workflow_id_or_name)}), 403
-        operational_parameters = {}
-        if request.json:
-            operational_parameters = request.json.get('operational_parameters')
         if status == START:
+            operational_parameters = {} if not request.json else \
+                request.json.get('operational_parameters', {})
             return _start_workflow(workflow, operational_parameters)
+        elif status == DELETED:
+            all_runs = True if request.json.get('all_runs') == 1 else False
```
diegodelemos: I would go for `true/false` see [OpenAPI docs](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types) and [JSON data types](https://www.w3schools.com/js/js_json_datatypes.asp)... It will ease coding the UI and it will be more understandable.

reana_workflow_controller/rest.py
```diff
@@ -1526,3 +1528,55 @@ def _get_current_job_progress(workflow_id):
...
+        if all_runs:
+            to_be_deleted += Session.query(Workflow).\
+                filter_by(name=workflow.name).all()
+        for workflow in to_be_deleted:
+            if hard_delete:
+                _hard_delete_workflow(workflow)
+            else:
+                _soft_delete_workflow(workflow)
+
+        return jsonify({'message': 'Workflow successfully deleted',
```
diegodelemos: Could we just return the dict and let the view decide what to do with it? I know there are other places where this is being done, but I would rather not do it.

reana_workflow_controller/rest.py
diegodelemos: I would go for now with the simpler approach of removing it straightaway since we would have to put in place more infrastructure to perform cron jobs.

reana_workflow_controller/rest.py
```diff
@@ -1526,3 +1528,55 @@ def _get_current_job_progress(workflow_id):
...
+            'Workflow {0}.{1} cannot be deleted as it'
+            ' is currently running.'.
+            format(
+                workflow.name,
+                workflow.run_number))
+
+
+def _hard_delete_workflow(workflow):
+    """Remove workflow row from database."""
+    remove_workflow_workspace(workflow.get_workspace())
```
diegodelemos: I think that this whole operation should be performed asynchronously because when we start to have a lot of big files and large amounts of runs this is going to take a while...

reana_workflow_controller/rest.py
dinosk: Sure, which dict do you mean? I was thinking of returning a list of dicts in the case that more than one workflow was deleted. It could be done with just a field:
```
jsonify({'deleted_workflows': [w.1, w.2, w.3]})
```
Are you thinking something different?

reana_workflow_controller/rest.py
dinosk: We could do some testing and see at which size this starts to lag, and maybe in another sprint switch to async? I like the idea

reana_workflow_controller/rest.py
dinosk: Fixed

reana_workflow_controller/rest.py
diegodelemos: Yes, let's do it in the next sprint, it was good to gather some thoughts though :)

reana_workflow_controller/rest.py
diegodelemos: I was thinking more about returning always the list of deleted workflows, and then inside the view `set_workflow_status` we would format the user message.

tests/test_views.py
```diff
@@ -708,3 +711,110 @@ def test_get_workflow_logs_unauthorized(app, default_user,
...
+@pytest.mark.parametrize("hard_delete", [True, False])
+def test_delete_workflow(app,
+                         session,
+                         default_user,
+                         sample_yadage_workflow_in_db,
+                         state,
+                         hard_delete):
+    """Test deletion of a workflow in all possible states."""
+    sample_yadage_workflow_in_db.status = state
+    session.add(sample_yadage_workflow_in_db)
```
diegodelemos: @dinosk are you getting any advantage from having `sample_yadage_workflow_in_db` to be a yadage workflow? I was reading the fixture documentation and [it states that it is a serial workflow](https://github.com/reanahub/pytest-reana/blob/master/pytest_reana/fixtures.py#L435).. would it be ok if I change it to be serial? either way a fix in `pytest-reana` has to be applied.

dinosk: I used the yadage specification which was there so that would be the advantage. Why not add another one instead of changing this? I can imagine we would need both in the future. Thanks for pointing out the docstring, to be fixed. The fixture works as it is here, I saw that it was added in plugin.py

reana_workflow_controller/rest.py
diegodelemos: Maybe you have this change locally and you didn't push it yet?

reana_workflow_controller/rest.py
```diff
@@ -1234,6 +1233,12 @@ def set_workflow_status(workflow_id_or_name):  # noqa
         if status == START:
             return _start_workflow(workflow,
                                    parameters)
+        elif status == DELETED:
+            all_runs = True if request.json.get('all_runs') else False
```
diegodelemos: We are missing the documentation of this options on the OpenAPI spec `docstring`.

dinosk: Description added in https://github.com/reanahub/reana-workflow-controller/pull/141/commits/efca3e419311bd5eb2f36c74aa3679a65f3f9e66#diff-3ee9faf1efe690830f3d86cb5d3cf590R1130

tests/test_views.py
```diff
@@ -740,3 +743,151 @@ def test_start_input_parameters(app, session, default_user,
...
+    session.commit()
+
+    _delete_workflow(sample_yadage_workflow_in_db, hard_delete=hard_delete)
```
diegodelemos: Should we use the REST API to do this instead of calling directly `_delete_workflow`? (in this and all the following tests). So we test our view logic and make sure that status responses and messages are correct? It is a place that we could "break", specially since we are aggregating all status operations in a single endpoint.

dinosk: @diegodelemos please take a look

tests/test_utils.py
```diff
@@ -0,0 +1,177 @@
...
+                                   yadage_workflow_with_name)
+from reana_db.models import Job, JobCache, Workflow, WorkflowStatus
+from werkzeug.utils import secure_filename
+
+from reana_workflow_controller.errors import WorkflowDeletionError
+from reana_workflow_controller.rest import START, STOP, _delete_workflow
+from reana_workflow_controller.utils import create_workflow_workspace
+
+
+@pytest.mark.parametrize("state", [WorkflowStatus.created,
```
diegodelemos: Should we use `status` instead of `state`? as we do everywhere else? (sorry I skipped at first)

reana_workflow_controller/rest.py
dinosk: I think we should do this change, and it will be more obvious when the utility functions move to their own utils.py, but I would not include it in this PR as it is currently in a consistent state (with the improvement pending, but still consistent). In the refactoring of the utils, we can have every non view method return the right response.

reana_workflow_controller/rest.py
dinosk: True, added the schema in commons: https://github.com/reanahub/reana-commons/pull/67

tests/conftest.py
```diff
@@ -30,21 +28,9 @@ def base_app(tmp_shared_volume_path):
         "SECRET_KEY": "SECRET_KEY",
         "TESTING": True,
         "SHARED_VOLUME_PATH": tmp_shared_volume_path,
-        "SQLALCHEMY_DATABASE_URI": "sqlite:///",
+        "SQLALCHEMY_DATABASE_URI": "sqlite:///testdb.db",
```
diegodelemos: @dinosk shall we create an issue about this?

PR #144, opened 2018-11-08
api: workflow parameters

roksys:
* Introduces new endpoint which returns workflow parameters.
   Connects https://github.com/reanahub/reana-client/issues/200

### DIFF
--- a/reana_workflow_controller/rest.py
+++ b/reana_workflow_controller/rest.py
@@ -192,9 +192,9 @@ def create_workflow():  # noqa
           schema:
             type: object
             properties:
-              operational_parameters:
+              operational_options:
                 type: object
-                description: Operational parameters.
+                description: Operational options.
               reana_specification:
                 type: object
                 description: >-
@@ -204,7 +204,7 @@ def create_workflow():  # noqa
                 description: Workflow name. If empty name will be generated.
             required: [reana_specification,
                        workflow_name,
-                       operational_parameters]
+                       operational_options]
       responses:
         201:
           description: >-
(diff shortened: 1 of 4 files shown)


reana_workflow_controller/rest.py
```diff
@@ -1526,3 +1644,13 @@ def _get_current_job_progress(workflow_id):
                 'prettified_cmd': job.prettified_cmd,
                 'current_job_name': job.name}
     return current_job_commands
+
+
+def _get_workflow_input_parameters(workflow):
```
diegodelemos: Shall we move this logic to the `Workflow` model on a new method?

reana_workflow_controller/rest.py
```diff
@@ -1526,3 +1644,13 @@ def _get_current_job_progress(workflow_id):
...
+    """Return workflow input parameters."""
+    # Apply live workflow parameters if given.
```
diegodelemos: Let's always move any comment to docstrings.

tests/test_views.py
```diff
@@ -708,3 +708,39 @@ def test_get_workflow_logs_unauthorized(app, default_user,
...
         assert res.status_code == 403
+
+
+def test_start_input_parameters(app, session, default_user,
+                                yadage_workflow_with_name):
+    """Test start workflow with inupt parameters."""
+    with app.test_client() as client:
+        os.environ["TESTS"] = "True"
+        # create workflow
+        res = client.post(url_for('api.create_workflow'),
```
diegodelemos: You can use an already existing fixture like `sample_yadage_workflow_in_db` instead of doing all the creation through the view. Example usage [here](https://github.com/reanahub/reana-workflow-controller/pull/147/files#diff-a4f5c4d64d4811b15e1405573c00f3f4R551).

reana_workflow_controller/rest.py
roksys: IMO this function is too specific to be in models and could cause confusion. Currently it is used only by serial. Once we have option to override input parameters for the other engines, we could move it to models. What do you think?

diegodelemos: Agreed! All ready to test run?

PR #146, opened 2018-11-13
rest: addition of workflow diff

dinosk:
* Adds get_workflow_diff method to return
  differences between the assets of two workflows.

Connects https://github.com/reanahub/reana-client/issues/193
Depends on https://github.com/reanahub/pytest-reana/pull/35

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -264,6 +264,108 @@
         "summary": "Create workflow and its workspace."
       }
     },
+    "/api/workflows/{workflow_id_or_name_a}/diff/{workflow_id_or_name_b}": {
+      "get": {
+        "description": "This resource shows the differences between the assets of two workflows. Resource is expecting two workflow UUIDs or names.",
+        "operationId": "get_workflow_diff",
+        "parameters": [
+          {
+            "description": "Required. UUID of workflow owner.",
+            "in": "query",
+            "name": "user",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "description": "Required. Analysis UUID or name of the first workflow.",
+            "in": "path",
+            "name": "workflow_id_or_name_a",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "description": "Required. Analysis UUID or name of the second workflow.",
+            "in": "path",
+            "name": "workflow_id_or_name_b",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "default": false,
+            "description": "Optional flag. If set, file contents are examined.",
+            "in": "query",
+            "name": "brief",
+            "required": false,
+            "type": "boolean"
+          },
+          {
+            "default": "5",
+            "description": "Optional parameter. Sets number of context lines for workspace diff output.",
+            "in": "query",
+            "name": "context_lines",
+            "required": false,
+            "type": "string"
+          }
+        ],
+        "produces": [
+          "application/json"
+        ],
+        "responses": {
+          "200": {
+            "description": "Request succeeded. Info about a workflow, including the status is returned.",
+            "examples": {
+              "application/json": {
+                "reana_specification": [
+                  "- nevents: 100000\n+ nevents: 200000"
+                ],
+                "workspace_listing": {
+                  "Only in workspace a: code": null
+                }
+              }
+            },
+            "schema": {
+              "properties": {
+                "reana_specification": {
+                  "type": "string"
+                },
+                "workspace_listing": {
+                  "type": "string"
+                }
+              },
+              "type": "object"
+            }
+          },
+          "400": {
+            "description": "Request failed. The incoming payload seems malformed.",
+            "examples": {
+              "application/json": {
+                "message": "Malformed request."
+              }
+            }
+          },
+          "403": {
+            "description": "Request failed. User is not allowed to access workflow.",
+            "examples": {
+              "application/json": {
+                "message": "User 00000000-0000-0000-0000-000000000000 is not allowed to access workflow 256b25f4-4cfb-4684-b7a8-73872ef455a1"
+              }
+            }
+          },
+          "404": {
+            "description": "Request failed. Either user or workflow does not exist.",
+            "examples": {
+              "application/json": {
+                "message": "Workflow 256b25f4-4cfb-4684-b7a8-73872ef455a1 does not exist."
+              }
+            }
+          },
+          "500": {
+            "description": "Request failed. Internal controller error."
+          }
+        },
+        "summary": "Get diff between two workflows."
+      }
+    },
     "/api/workflows/{workflow_id_or_name}/logs": {
       "get": {
         "description": "This resource is expecting a workflow UUID and a filename to return its outputs.",
@@ -537,7 +639,8 @@
             "description": "Required. New status.",
             "enum": [
               "start",
-              "stop"
+              "stop",
+              "deleted"
             ],
             "in": "query",
             "name": "status",
--- a/reana_workflow_controller/rest.py
+++ b/reana_workflow_controller/rest.py
@@ -44,10 +44,9 @@ from reana_workflow_controller.utils import (create_workflow_workspace,
                                              remove_workflow_workspace)
 
 START = 'start'
-STOP = 'stopped'
-PAUSE = 'pause'
+STOP = 'stop'
 DELETED = 'deleted'
-STATUSES = {START, STOP, PAUSE, DELETED}
+STATUSES = {START, STOP, DELETED}
 
 restapi_blueprint = Blueprint('api', __name__)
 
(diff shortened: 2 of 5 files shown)


reana_workflow_controller/rest.py
```diff
@@ -1485,6 +1491,134 @@ def get_workflow_parameters(workflow_id_or_name):  # noqa
...
+              }
+        500:
+          description: >-
+            Request failed. Internal controller error.
+    """
+    try:
+        user_uuid = request.args['user']
+        brief = request.args.get('brief', False)
+        brief = True if brief == 'true' else False
+        context_lines = request.args.get('context_lines', 5)
```
diegodelemos: We can document the defaults for `brief` and `context_lines`.

reana_workflow_controller/rest.py
```diff
@@ -1485,6 +1491,134 @@ def get_workflow_parameters(workflow_id_or_name):  # noqa
...
+            workflow_a, workflow_b)
+        workspace_diff = get_workspace_diff(
+            workflow_a, workflow_b, brief, context_lines)
+        response = {'reana_specification': json.dumps(specification_diff),
+                    'workspace_listing': json.dumps(workspace_diff)}
+        return jsonify(response)
+    except REANAWorkflowControllerError as e:
+        return jsonify({"message": str(e)}), 409
+    except WorkflowInexistentError as e:
+        wrong_workflow = workflow_id_or_name_b if workflow_a_exists \
```
diegodelemos: Why do we need this auxiliary variable? Wouldn't it be enough to check if `workflow_id_or_name_a` is `None`?

reana_workflow_controller/rest.py
```diff
@@ -1861,3 +1995,91 @@ def _mark_workflow_as_deleted_in_db(workflow):
...
+
+def get_specification_diff(workflow_a, workflow_b, output_format=None):
+    """Return differences between two workflow specifications.
+
+    :param workflow_a: The first workflow to be compared.
+    :type: reana_db.models.Workflow instance.
+    :param workflow_a: The first workflow to be compared.
+    :type: reana_db.models.Workflow instance.
+    :param output_format: Sets output format. Optional.
+    :type: String. One of ['unified', 'context', 'html'].
```
diegodelemos: Why do we set to `None` the default in the signature and later we say that it will be `unified` by default? If we default to `unified` we do not need `None` around, do we?

reana_workflow_controller/rest.py
```diff
@@ -1861,3 +1995,91 @@ def _mark_workflow_as_deleted_in_db(workflow):
...
+
+    :rtype: List with lines of differences.
+    """
+    if output_format not in ['unified', 'context', 'html', None]:
+        raise ValueError('Unknown output format.'
+                         'Please select one of unified, context or html.')
+
+    if output_format in [None, 'unified']:
+        diff_method = getattr(difflib, 'unified_diff')
+        # specification_diff = list(difflib.unified_diff(spec_a, spec_b))[2:]
```
diegodelemos: If we don't need this code anymore we can remove the comments.

reana_workflow_controller/rest.py
```diff
@@ -1861,3 +1995,91 @@ def _mark_workflow_as_deleted_in_db(workflow):
...
+        # specification_diff = difflib.HtmlDiff(spec_a, spec_b)
+    specification_diff = dict.fromkeys(workflow_a.reana_specification.keys())
+    for section in specification_diff:
+        section_a = pprint.pformat(
+            workflow_a.reana_specification.get(section, '')).\
+            splitlines()
+        section_b = pprint.pformat(
+            workflow_b.reana_specification.get(section, '')).\
+            splitlines()
+        specification_diff[section] = list(diff_method(section_a,
```
diegodelemos: Maybe a comment here explaining why we start from `2` wouldn't hurt.

reana_workflow_controller/rest.py
```diff
@@ -1861,3 +1995,91 @@ def _mark_workflow_as_deleted_in_db(workflow):
     current_db_sessions = Session.object_session(workflow)
     current_db_sessions.add(workflow)
     current_db_sessions.commit()
+
+
+def get_specification_diff(workflow_a, workflow_b, output_format=None):
+    """Return differences between two workflow specifications.
+
+    :param workflow_a: The first workflow to be compared.
+    :type: reana_db.models.Workflow instance.
```
diegodelemos: Should we also add an `intersphinx` link?

reana_workflow_controller/rest.py
```diff
@@ -1861,3 +1995,91 @@ def _mark_workflow_as_deleted_in_db(workflow):
...
+    :param workflow_b: The second workflow to be compared.
+    :type: reana_db.models.Workflow instance.
+    :param brief: Optional flag to show brief workspace diff.
+    :type: Boolean.
+    :param context_lines: The number of context lines to show above and after
+                          the discovered differences.
+    :type: Integer or string.
+
+    :rtype: Dictionary with file paths and their sizes
```
diegodelemos: This would be `:return` and `:rtype` dict, no? @reanahub/developers are we happy to start adding type annotations? Since here we are running on Python 3 on all cluster components I would rather start adding Python 3 type annotations than this ones.

tests/test_views.py
```diff
@@ -980,3 +982,75 @@ def test_deletion_of_workspace_of_an_already_deleted_workflow(
...
+                            tmp_shared_volume_path,
+                            sample_workflow_workspace):
+    """Test get workspace differences."""
+    # create the workspaces for the two workflows
+    workspace_path_a = next(sample_workflow_workspace(
+        str(sample_serial_workflow_in_db.id_)))
+    workspace_path_b = next(sample_workflow_workspace(
+        str(sample_yadage_workflow_in_db.id_)))
+
+    sample_serial_workflow_in_db.get_workspace = lambda: str(
```
diegodelemos: Why do we do this?

tests/test_views.py
```diff
@@ -980,3 +982,75 @@ def test_deletion_of_workspace_of_an_already_deleted_workflow(
...
+        sample_serial_workflow_in_db.id_)
+    sample_yadage_workflow_in_db.get_workspace = lambda: str(
+        sample_yadage_workflow_in_db.id_)
+    # modify the contents in one file
+    with open(
+            os.path.join(
+                workspace_path_a, 'data',
+                'World_historical_and_predicted_populations_in_percentage.csv'
```
diegodelemos: I would keep the test file names simple and short.

diegodelemos:
Needs rebase.

reana_workflow_controller/rest.py
dinosk: No need but it makes for better readability

tests/test_views.py
dinosk: This is using the same as in world population example, could be changed in `pytest-reana`

reana_workflow_controller/rest.py
dinosk: This follows the docstring styles in the utility methods in `rest.py`

docs/openapi.json
```diff
@@ -537,7 +637,8 @@
             "description": "Required. New status.",
             "enum": [
               "start",
-              "stop"
+              "stop",
+              "deleted"
```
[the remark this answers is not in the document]
diegodelemos: I understand, but I just realised now, it was just trying to avoid having to wait for a new set of PRs to be integrated for a string change, but I will create then now no problem.

tests/test_views.py
dinosk: This is interesting, in `rest.py` `workflow_instance.get_workspace()` will return a path like:
```
'/reana/users/user-id/workflows/workflow-id'
```
where in the tests the root directory is `/tmp/pytest-of-user/....` so it needs to be patched.

diegodelemos: mmm I am confused, [`get_workspace` always returns a relative path](https://github.com/reanahub/reana-db/blob/master/reana_db/utils.py#L13), then in the Flask app we [contextualize it to the `SHARED_VOLUME_PATH` value](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/rest.py#L377),  is it maybe the config variable that is not being correctly set?

dinosk: As you can see by running this test, `SHARED_VOLUME_PATH` is set correctly:
```
ipdb> workspace_a
'users/00000000-0000-0000-0000-000000000000/workflows/3df8c716-8fba-4fb5-ab7d-2d52500064e9'
ipdb> current_app.config['SHARED_VOLUME_PATH']
'/tmp/pytest-of-dinos/pytest-68/reana0'
```

dinosk:
@diegodelemos thank you for the review, but I have to say that I personally do not appreciate the tone of this review. It might be good to discuss how to conduct code reviews with @reanahub/developers

PR #147, opened 2018-11-13
tasks: add stop option

diegodelemos:
* Finalises workflow execution through revoke method (which is going
  to kill the process in which the task was running).

Connect reanahub/reana-client#194.
Blocked by https://github.com/reanahub/pytest-reana/pull/26.

### DIFF
--- a/reana_workflow_controller/rest.py
+++ b/reana_workflow_controller/rest.py
@@ -30,14 +30,15 @@ from reana_workflow_controller.errors import (REANAWorkflowControllerError,
                                               WorkflowNameError)
 from reana_workflow_controller.tasks import (run_cwl_workflow,
                                              run_serial_workflow,
-                                             run_yadage_workflow)
+                                             run_yadage_workflow,
+                                             stop_workflow)
 from reana_workflow_controller.utils import (create_workflow_workspace,
                                              list_directory_files,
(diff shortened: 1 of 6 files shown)


diegodelemos:
Test failing because of https://github.com/reanahub/pytest-reana/pull/32.

reana_workflow_controller/rest.py
```diff
@@ -1417,6 +1428,36 @@ def _start_workflow(workflow, parameters):
...
+                queue=WORKFLOW_QUEUES[workflow.type_],
+                task_id='delete-{}'.format(str(workflow.id_))
+            )
+        else:
+            raise NotImplementedError(
+                'Workflow type {} does not support '
+                'stop yet.'.format(workflow.type_))
+    else:
+        message = \
+            ("Workflow {id_} could not be stopped. Only running workflows"
```
dinosk: reana-client has the same first line in the error message https://github.com/reanahub/reana-client/pull/205/files#diff-4f27f08c3f46f0b755c1d5ac1a1e2dcaR445 and stopping a not running workflow shows:
```
Workflow could not be stopped: 
Workflow 7044131f-bc73-455d-94aa-a56c83cf8677 could not be stopped. Only running workflows can be stopped.
```
This message could be sth like "Workflow {id_} is not running."

Issue #158, opened 2018-12-06
workflow-execution: create and implement interface to start RWE_'s


diegodelemos:
Since [workflow engines will stop being Celery tasks](https://github.com/reanahub/reana-workflow-engine-cwl/issues/75) we change the api to start workflows. Now, we should define an interface which can support the following actions on workflows:

| WorkflowExecutionManager              |
|------------------------------------ |
| # start_batch_workflow_run               |
| # start_interactive_workflow_run       |
| # stop_workflow_run                           |
| ...                                                          |

And we should create the Kubernetes implementation, i.e. `KubernetesWorkflowExecutionManager`.
For now we will only implement the method `start_batch_workflow`. `start_interactive_workflow` will be implemented later and `stop_workflow` will be implemented in https://github.com/reanahub/reana-workflow-controller/issues/157.

PR #159, opened 2018-12-07
workflow-execution: start RWE_ on Kubernetes

diegodelemos:
Connect #158.
Depends on https://github.com/reanahub/reana-commons/pull/75.


reana_workflow_controller/config.py
```diff
@@ -50,3 +50,10 @@
 WORKFLOW_QUEUES = {'cwl': 'cwl-default-queue',
                    'yadage': 'yadage-default-queue',
                    'serial': 'serial-default-queue'}
+
+WORKFLOW_ENGINE_SHARED_VOLUME = 'vm'
```
tiborsimko: Hmm, `vm` may be perhaps confusion-prone, what about using `local`? (or perhaps `host`?)

diegodelemos: Thanks! I think host is the most representative, I was not sure what to use here 😕

reana_workflow_controller/config.py
```diff
@@ -50,3 +53,23 @@
 WORKFLOW_QUEUES = {'cwl': 'cwl-default-queue',
                    'yadage': 'yadage-default-queue',
                    'serial': 'serial-default-queue'}
+
+REANA_STORAGE_BACKEND = os.getenv('REANA_STORAGE_BACKEND', 'local')
+"""Type of storage attached to the engines, one of ['local', 'ceph']."""
+
+MANILA_CEPHFS_PVC = 'manila-cephfs-pvc'
+"""If CEPH storage backned is used, this represents the name of the
```
okraskaj: typo in the docstring:
ba**ckn**ed where it should be ba**cke**nd

reana_workflow_controller/workflow_run_managers/kubernetes/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+            "hostPath": {
+                "path": SHARED_FS_MAPPING['MOUNT_SOURCE_PATH'],
+            }
+        }
+    }
+    """Configuration to connect to the different storage backends."""
+
+    def start_batch_workflow_run(self):
+        """Start a batch workflow run."""
+        namespace = "default"
```
okraskaj: I think here we should have single quotes

diegodelemos: Regarding this, there is already a mix in the code base what do you think about using something like https://github.com/ambv/black? @reanahub/developers

tiborsimko: "Black prefers double quotes over single quotes" :wink:

PR #161, opened 2018-12-17
workflow execution: pass arguments to rwe_ individual runs

okraskaj:
Connects reanahub/reana-workflow-engine-serial#5

### DIFF
--- a/reana_workflow_controller/workflow_run_managers/__init__.py
+++ b/reana_workflow_controller/workflow_run_managers/__init__.py
@@ -5,6 +5,7 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 
 """Workflow run manager interface."""
+import json
 
 from reana_workflow_controller.config import WORKFLOW_ENGINE_VERSION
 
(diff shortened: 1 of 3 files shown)


reana_workflow_controller/workflow_run_managers/__init__.py
```diff
@@ -27,15 +28,20 @@ class WorkflowRunManager():
...
+                'environment_variables': common_env_variables},
         'serial': {'image': 'reanahub/reana-workflow-engine-serial:{}'.format(
             WORKFLOW_ENGINE_VERSION),
-                   'command': ['sleep', '1000'],
+                   'command': ("run-serial-workflow "
+                               "--workflow-uuid {id} "
+                               "--workflow-workspace {workspace} "
+                               "--workflow-json '{json}' "
+                               "--workflow-parameters '{parameters}' "
+                               "--operational-options {options} "),
```
tiborsimko: Quote `options` like `parameters`?

(we can revise later whether this is safe regarding escaping/unescaping... there are ways to encode/decode inputs for example)

reana_workflow_controller/workflow_run_managers/__init__.py
```diff
@@ -81,7 +87,13 @@ def _workflow_engine_image(self):
     def _workflow_engine_command(self):
         """Return the command to be run for a given workflow engine."""
         return (WorkflowRunManager.engine_mapping[self.workflow.type_]
-                ['command'])
+                ['command'].format(
+             id=self.workflow.id_,
+             workspace=self.workflow.get_workspace(),
+             json=json.dumps(self.workflow.get_specification()),
+             parameters=json.dumps(self.workflow.get_input_parameters()),
```
roksys: `(self.workflow.get_input_parameters()` returns only original parameters, without applied live parameters (which are provided `$ reana-client run -p region=Europe`). 

One possible solution is 

`parameters=json.dumps(self._get_workflow_input_parameters()),`

and add this function 

```
    def _get_workflow_input_parameters(self):
        """Return workflow input parameters merged with live ones, if given."""
        if self.workflow.input_parameters:
            return dict(self.workflow.get_input_parameters(),
                        **self.workflow.input_parameters)
        else:
            return self.workflow.get_input_parameters()
```

I am not sure if this function should be here or in models.

tiborsimko: For the time being such a "get merged workflow input parameters" function can simply live here. We can then take time to think about harmonising naming everywhere to distinguish between "workflow default parameters" (taken from `reana.yaml`) and "workflow command-line parameters" (taken from CLI, overriding the default ones) with the final "workflow input parameters" being the merge of the two. We could then have one single getter function and distinguish context if need be e.g. `get_workflow_parameters(name, context='spec|cli')` which would return "merged" values by default, but could also return one or the other input channel contexts, if need be.  This is just an example for illustration... precise names can differ based on what we store in DB as split or combined. We can deal with this in another ticket, and if we do, then such a function should probably go to `r-commons` indeed, as more components could start taking advantage of it.

PR #165, opened 2019-01-09
global: serverless CWL workflow engine

roksys:
* Allows to create CWL workflow engine on demand.
  Connects reanahub/reana-workflow-engine-cwl/issues/75


reana_workflow_controller/workflow_run_managers/__init__.py
```diff
@@ -21,14 +21,27 @@ class WorkflowRunManager():
         {
             'name': 'SHARED_VOLUME_PATH',
             'value': '/reana'
+        },
+        {
+            'name': 'WDB_SOCKET_SERVER',
```
tiborsimko: Note that the WDB part needs to be removed here, as it will be properly if-else-d in another PR...

PR #170, opened 2019-01-17
config: set SHARED_VOLUME_PATH to /var/reana

dinosk:
Connects https://github.com/reanahub/reana-cluster/issues/131 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,8 +14,8 @@ COPY CHANGES.rst README.rst setup.py /code/
 COPY reana_workflow_controller/version.py /code/reana_workflow_controller/
 
 WORKDIR /code
-RUN pip install --no-cache-dir requirements-builder && \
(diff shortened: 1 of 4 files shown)


diegodelemos:
There is a problem with the Docker builds which I can not reproduce locally. Same in https://github.com/reanahub/reana-workflow-engine-yadage/pull/97, https://github.com/reanahub/reana-workflow-controller/pull/170 and https://github.com/reanahub/reana-workflow-engine-cwl/pull/81.

dinosk:
Problem occurs because of `--no-cache-dir` in the Dockerfiles: https://github.com/reanahub/reana-workflow-controller/blob/master/Dockerfile#L17

It is marked as a TODO in pip: https://github.com/pypa/pip/blame/master/src/pip/_internal/wheel.py#L843

Issue #160, opened 2018-12-14
installation: workflow controller only runs when installed with editable mode


diegodelemos:
Working on [current master](https://github.com/reanahub/reana-workflow-controller/commit/387041409bd16ba7b430599304a10c8571a94ba3):
```console
$ pip install -e .
$ export FLASK_APP=reana_workflow_controller/app.py
$ flask run --host=0.0.0.0
 * Serving Flask app "reana_workflow_controller/app.py"
 * Environment: production
   WARNING: Do not use the development server in a production environment.
   Use a production WSGI server instead.
 * Debug mode: off
 * Running on http://0.0.0.0:5000/ (Press CTRL+C to quit)
$ pip install .
$ flask run --host=0.0.0.0
 * Serving Flask app "reana_workflow_controller/app.py"
 * Environment: production
   WARNING: Do not use the development server in a production environment.
   Use a production WSGI server instead.
 * Debug mode: off
Usage: flask run [OPTIONS]

Error: While importing "reana_workflow_controller.app", an ImportError was raised:

Traceback (most recent call last):
  File "/Users/rodrigdi/.virtualenvs/r-w-controller/lib/python3.6/site-packages/flask/cli.py", line 235, in locate_app
    __import__(module_name)
  File "/Users/rodrigdi/.virtualenvs/r-w-controller/lib/python3.6/site-packages/reana_workflow_controller/app.py", line 17, in <module>
    app = create_app()
  File "/Users/rodrigdi/.virtualenvs/r-w-controller/lib/python3.6/site-packages/reana_workflow_controller/factory.py", line 28, in create_app
    from .rest import restapi_blueprint  # noqa
  File "/Users/rodrigdi/.virtualenvs/r-w-controller/lib/python3.6/site-packages/reana_workflow_controller/rest.py", line 45, in <module>
    from reana_workflow_controller.workflow_run_managers.kubernetes import \
ModuleNotFoundError: No module named 'reana_workflow_controller.workflow_run_managers'
```
This has not been spotted because of running on [DEBUG mode](https://github.com/reanahub/reana-workflow-controller/blob/master/Dockerfile#L25) with `docker build --build-arg DEBUG=true . -t reanahub/reana-workflow-controller:latest`.

tiborsimko:
BTW looking at the file structure, the code seems to live mostly inside `__init__.py` files:

```console
$ find reana_workflow_controller/workflow_run_managers  
reana_workflow_controller/workflow_run_managers
[...]

PR #171, opened 2019-01-17
workflow-execution: reimplement workflow stop

### LINKED ISSUE #157: workflow-execution: reimplement workflow stop
Currently, [the workflow stop feature](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/rest.py#L1681) is implemented using an [auxiliary celery task](https://github.com/reanahub/reana-commons/blob/master/reana_commons/tasks.py#L22)

On a serverless workflows architecture we won't have Celery service per engine, so we have two options:

- Introduce a REANA-Worker ([work is already started to have a REANA-Worker](https://github.com/reanahub/reana-cluster/pull/91)) who will use [the mentioned task](https://github.com/reanahub/reana-commons/blob/master/reana_commons/tasks.py#L22) and it will: kill the dependent jobs as it is done right now, and reimplement the part of killing the workflow (`revoke(workflow_uuid, terminate=True)`) using Kubernetes API instead.
- Since we already depend on Kubernetes, implement stop within the RWE_ itself, and take advantage of [Kubernetes lifecycle hooks](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks), using a `PreStop` hook which would do the clean-up.

Depends on https://github.com/reanahub/reana-workflow-controller/issues/158.

diegodelemos:
* (closes #157).
* Addresses https://github.com/reanahub/reana-workflow-controller/issues/160#issuecomment-447789462.


reana_workflow_controller/workflow_run_manager.py
```diff
@@ -125,3 +137,100 @@ def _workflow_engine_env_vars(self):
...
+
+    def start_batch_workflow_run(self):
+        """Start a batch workflow run."""
+        workflow_run_name = self._workflow_run_name_generator('batch')
+        job = self._create_job_spec(workflow_run_name)
+        current_k8s_batchv1_api_client.create_namespaced_job(
+            namespace=KubernetesWorkflowRunManager.default_namespace,
+            body=job)
+
+    def stop_batch_workflow_run(self, workflow_run_jobs=None):
```
diegodelemos: Just stop has been added, the rest has been moved from the old Kubernetes file.

PR #174, opened 2019-01-23
workflow-execution: configurable engine versions

diegodelemos:
Connects https://github.com/reanahub/reana-cluster/issues/142.

### DIFF
--- a/reana_workflow_controller/config.py
+++ b/reana_workflow_controller/config.py
@@ -67,8 +67,17 @@ SHARED_FS_MAPPING = {
 }
 """Mapping from the shared file system backend to the job file system."""
 
-WORKFLOW_ENGINE_VERSION = 'cvmfs-16'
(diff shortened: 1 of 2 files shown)


reana_workflow_controller/config.py
```diff
@@ -53,7 +53,7 @@
 """If CEPH storage backend is used, this represents the name of the
 Kubernetes persistent volume claim."""
 
-MOUNT_CVMFS = os.getenv('REANA_MOUNT_CVMFS', False)
+MOUNT_CVMFS = True if os.getenv('REANA_MOUNT_CVMFS') == 'True' else False
```
tiborsimko: Maybe add a `lower()` to catch possible values of "true" as well as "True"?  

(In the shell context, when we go out of Python, it may be safer to use otherwise-named strings (e.g. `CACHE=on`) or numbers (e.g.  `FLASK_DEBUG=1`) instead of "true/false", precisely helping to avoid possible type casting troubles. Anyway, we can keep it like this, just thinking out loud...)

PR #179, opened 2019-02-04
interactive-execution: first implementation

### LINKED ISSUE #178: interactive-execution: run interactive web notebooks
Currently, REANA only supports running [batch workflows](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/workflow_run_manager.py#L168). We want now to add the functionality to run interactive notebooks, meaning that users would be able to attach to a notebook running inside REANA and be able to use its computing power remotely. To implement that we need:

- [x] Install Traefik on Minikube (R-Cluster)
- [x] Implement the creation of interactive pods (R-W-Controller)
- [x] Implement the logic to expose the dynamically created pods (R-W-Controller)
- [ ] Protect interactive session, an option with Jupyter is to return [the autogenerated token](https://jupyter-notebook.readthedocs.io/en/stable/security.html#security-in-the-jupyter-notebook-server). It would be great if we could configure the Notebook to instead of autogenerating the token, to use the REANA user token.
- [ ] Solve redirect issue see https://github.com/reanahub/reana-workflow-controller/issues/178#issuecomment-461341399.
- [ ] Mount workflow workspace inside the interactive session pod.
- [ ] Provide a way to stop interactive sessions (e.g. `reana-client close workflow.1`)
- [ ] Gargabe collect all intactive interactive sessions (overall execution time policy, `cronjob` maybe, and no traffic policy for example, some native garbage collector similar to [TTL mechanism for finished jobs](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#ttl-mechanism-for-finished-jobs))
- [ ] REANA Cluster needs rewrite to retrieve the API Server URL, connects with https://github.com/reanahub/reana-cluster/issues/73.

diegodelemos:
* (closes #178).


reana_workflow_controller/rest.py
```diff
@@ -1129,6 +1129,108 @@ def set_workflow_status(workflow_id_or_name):  # noqa
...
+            type: object
+            properties:
+              image:
+                type: string
+              port:
+                type: integer
+      responses:
+        200:
+          description: >-
+            Request succeeded. The file has been added to the workspace.
```
roksys: Description from other function

reana_workflow_controller/rest.py
```diff
@@ -1129,6 +1129,108 @@ def set_workflow_status(workflow_id_or_name):  # noqa
...
+          description: >-
+            Request failed. Internal controller error.
+    """
+    try:
+        if request.json and not request.json.get("image"):
+            raise ValueError("If interactive_environment payload is sent, it˛"
+                             "should contain the image property.")
+
+        user_uuid = request.args["user"]
+        image = request.json.get("image", None)
```
roksys: Why not setting to default ones from reana_workflow_controller/config.py, instead of None?

reana_workflow_controller/rest.py
```diff
@@ -1129,6 +1129,108 @@ def set_workflow_status(workflow_id_or_name):  # noqa
...
+        port = request.json.get("port", None)
+        workflow = None
```
roksys: is this needed?

diegodelemos: It is to know in the exception (`except (KeyError, ValueError) as e:`) if it has been a 404 or a 400. Inside `_get_workflow_with_uuid_or_name` we are throwing a `ValueError` when workflow doesn't exist  so that was the way to patch it.. we should fix `_get_workflow_with_uuid_or_name` to throw a proper exception so we know what happened.

PR #195, opened 2019-03-06
interactive-execution: add REANA shared storage

### LINKED ISSUE #186: interactive-execution: mount workflow workspace on interactive session pod
As we are doing for the [batch workflow engines](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/workflow_run_manager.py#L294-L297), we have to mount the workflow workspace on the interactive session pods (missing [here](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/templates/k8s/interactive-deployment.template)). This way users that log in into the interactive session will be able to manage the workflow workspace and perform new exploratory work with it.

While implementing it we should discuss about moving the current way of building the interactive template as [string template](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/templates/k8s/interactive-deployment.template) to use the Kubernetes Python library as we do for the [job spec creation](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/workflow_run_manager.py#L254).

diegodelemos:
Closes #186.

### DIFF
--- a/reana_workflow_controller/utils.py
+++ b/reana_workflow_controller/utils.py
@@ -25,8 +25,8 @@ def create_workflow_workspace(path):
     :return: Absolute workspace path.
     """
     reana_fs = fs.open_fs(app.config['SHARED_VOLUME_PATH'])
-    if not reana_fs.exists(path):
-        reana_fs.makedirs(path)
+    reana_fs.makedirs(path, recreate=True)
+    os.chmod(reana_fs.getsyspath(path), mode=0o774)
 
 
 def list_directory_files(directory):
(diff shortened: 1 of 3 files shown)


reana_workflow_controller/utils.py
```diff
@@ -11,6 +11,7 @@
 from pathlib import Path
 
 import fs
+from fs.permissions import Permissions
```
roksys: Seems to be imported but no used, right?

PR #196, opened 2019-03-06
k8s: dependent interactive objects

roksys:
* Makes k8s objects created for an interactive session dependent.
  If ingress object gets deleted, service and deployment objects
  are deleted automaticallyi.
  Closes reanahub/reana-client/issues/247

### DIFF
--- a/reana_workflow_controller/k8s.py
+++ b/reana_workflow_controller/k8s.py
@@ -179,12 +179,27 @@ def instantiate_k8s_objects(kubernetes_objects, namespace):
         current_k8s_extensions_v1beta1.create_namespaced_ingress
     }
     try:
-        for obj in kubernetes_objects.items():
+        parent_k8s_object_references = None
+        for index, obj in enumerate(kubernetes_objects.items()):
             kind = obj[0]
             k8s_object = obj[1]
-            instantiate_k8s_object[kind](namespace, k8s_object)
+            if index == 0:
+                result = instantiate_k8s_object[kind](namespace, k8s_object)
+                parent_k8s_object_references = [{
+                    "uid": result._metadata.uid,
+                    "kind": result._kind,
+                    "name": result._metadata.name,
+                    "apiVersion": result._api_version,
+                }]
+            else:
+                k8s_object.metadata.owner_references =  \
+                    parent_k8s_object_references
+                result = instantiate_k8s_object[kind](namespace, k8s_object)
     except KeyError:
         raise Exception("Unsupported Kubernetes object kind {}.".format(kind))
+    except ApiException as e:
+        raise Exception("Exception when calling ExtensionsV1beta1Api->"
+                        "create_namespaced_deployment_rollback: %s\n" % e)
 
 
 def delete_k8s_objects_if_exist(kubernetes_objects, namespace):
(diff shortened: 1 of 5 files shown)


reana_workflow_controller/k8s.py
```diff
@@ -179,12 +179,28 @@ def instantiate_k8s_objects(kubernetes_objects, namespace):
         current_k8s_extensions_v1beta1.create_namespaced_ingress
```
diegodelemos: What do you think about either renaming the function to `instantiate_chained_k8s_objects` or something similar which states this new behaviour or we make the "chaining" mechanism optional with an argument?

reana_workflow_controller/k8s.py
```diff
@@ -179,12 +179,28 @@ def instantiate_k8s_objects(kubernetes_objects, namespace):
...
     }
     try:
-        for obj in kubernetes_objects.items():
+        parent_k8s_object_references = None
+        for index, obj in enumerate(kubernetes_objects.items()):
             kind = obj[0]
             k8s_object = obj[1]
-            instantiate_k8s_object[kind](namespace, k8s_object)
+            if index == 0:
```
diegodelemos: Shall we remove this `if` by picking the first `kubernetes_objects.items()` element directly and then iterating over the rest (if they exist) assingning them the `owner_references`?

reana_workflow_controller/rest.py
```diff
@@ -17,27 +17,28 @@
 from datetime import datetime
 from uuid import UUID, uuid4
 
+from kubernetes.client.rest import ApiException
```
diegodelemos: Unused I think.

reana_workflow_controller/workflow_run_manager.py
```diff
@@ -126,6 +127,13 @@ def start_interactive_session(self):
         """Start an interactive workflow run."""
         raise NotImplementedError('')
 
+    def stop_interactive_session(self):
```
diegodelemos: Let's move this implementation to `KubernetesWorkflowRunManager` since it is Kubernetes specific and leave a not implemented here.

reana_workflow_controller/rest.py
```diff
@@ -1269,6 +1270,94 @@ def open_interactive_session(workflow_id_or_name, interactive_session_type):  #
...
+                                                   user_uuid)
+        if workflow.interactive_session_name is None:
+            return jsonify(
+                {"message": "Workflow - {} has no open interactive session."
+                            .format(workflow_id_or_name)}), 404
+        kwrm = KubernetesWorkflowRunManager(workflow)
+        kwrm.stop_interactive_session()
+        workflow.interactive_session_name = None
+        workflow.interactive_session = None
+        current_db_sessions = Session.object_session(workflow)
```
diegodelemos: What do you think about moving this DB business inside the `stop_interactive_session()` implementation? We are already doing something similar for opening, we set the interactive session name and path in DB inside the method.

reana_workflow_controller/k8s.py
roksys: I was thinking about that, but IMO `instantiate_chained_k8s_objects` gets much bigger and more complicated, because `kubernetes_objects` is a dict and it is impossible to loop starting from 2nd item `kubernetes_objects.items()[1:]`, I would need to do some conversions.

PR #209, opened 2019-04-17
release: 0.5.0

### LINKED ISSUE #208: global: run release checklist

### LINKED ISSUE #210: installation: dependency conflict during kubernetes installation
In kubernetes library there is a conflict (even when installing on a clean virtualenv):

``` requests 2.21.0 has requirement urllib3<1.25,>=1.21.1, but you'll have urllib3 1.25 which is incompatible. ```

it may have something to do with failing build of urllib3 https://travis-ci.org/urllib3/urllib3

Probably the solution is to limit the versions in the affected repositories as `requests` did:
```
'urllib3<1.25,>=1.21.1'
```

After the dependency conflict is solved we should unpin it

### PR DESCRIPTION
**Pull request by okraskaj**
* Closes #208
* Fix import sorting to uncomment isort in run_test.py.
* Fix docs badge link (Closes reanahub/reana#162)
* Fix kubernetes dependency conflict (Closes #210)

Signed-off-by: Jan Okraska <jan.okraska@cern.ch>

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,22 @@
 Changes
 =======
+Version 0.5.0 (2019-04-23)
+--------------------------
+
+- Adds support to create interactive sessions so the workspace can be explored
+  and modified through a Jupyter notebook.
+- Creates workflow engine instances on demand for each user and makes CVMFS
+  available inside of them.
+- Adds new endpoint to compare two workflows. The output is a ``git`` like
+  diff which can be configured to show differences at metadata level,
+  workspace level or both.
+- Adds new endpoint to delete workflows including the stopped ones.
+- Adds new endpoints to delete and move files whithin the workspace.
+  The deletion can be also done recursively with a wildcard.
+- Adds new endpoint which returns workflow parameters.
+- Adds new endpoint to query the disk usage of a given workspace.
+- Makes docker image slimmer by using ``python:3.6-slim``.
+- Centralises log level and log format configuration.
 
 Version 0.4.0 (2018-11-06)
 --------------------------
--- a/reana_workflow_controller/utils.py
+++ b/reana_workflow_controller/utils.py
@@ -10,12 +10,12 @@
 import os
 from pathlib import Path
 
-from reana_db.database import Session
-from reana_db.models import Job, JobCache
-
 import fs
 from flask import current_app as app
 from reana_commons.config import REANA_WORKFLOW_UMASK
+from reana_db.database import Session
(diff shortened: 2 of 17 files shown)

### CODE REVIEW

File: reana_workflow_controller/utils.py [hunk 1]
```diff
@@ -72,16 +72,20 @@ def remove_workflow_jobs_from_cache(workflow):
 
 
 def remove_files_recursive_wildcard(directory_path, path):
-    """Remove file(s) from workflow workspace.
-
-    :param directory_path: FIXME.
-    :param path: FIXME.
-    :return: FIXME.
+    """Remove file(s) fitting the wildcard from workflow workspace.
```
roksys: Now we try to use only ``workspace``, not ``workflow workspace``. CLI e.g https://github.com/reanahub/reana-client/blob/256420656f8201ad5d81295c357b6cf35725b681/docs/cmd_list.txt#L36

okraskaj: 👍 fixed

reana_workflow_controller/utils.py
```diff
@@ -72,16 +72,20 @@ def remove_workflow_jobs_from_cache(workflow):
...
+    :return: Dictionary with the results:
+       - dictionary with names of succesfully deleted files and their sizes
+       - dictionary with names of failed deletions and corresponding
+       error messages.
     """
     deleted = {"deleted": {}, "failed": {}}
     secure_path = remove_upper_level_references(path)
     posix_dir_prefix = Path(directory_path)
     paths = list(posix_dir_prefix.glob(secure_path))
+    # sort paths by length to start with the leafs of the directory tree
```
roksys: I think it should be ``leaves ``, right?

CHANGES.rst
```diff
@@ -1,6 +1,23 @@
 Changes
 =======
 
+Version 0.5.0 (2019-04-23)
+--------------------------
+
+- Enables to create workflows on demand.
+- Add support for interactive sessions to manage workflows directly with interactive tools such as Jupyter.
```
diegodelemos: What about: "- Adds support to create interactive sessions so the workspace can be explored and modified through a Jupyter notebook."?

CHANGES.rst
```diff
@@ -1,6 +1,23 @@
...
+- Adds REANA shared storage to interactive sessions for easier viewing of the results.
```
diegodelemos: I think this would be already included in the previous bullet point, since we mention that the workspace can be accessed by the interactive session.

PR #229, opened 2019-07-18
workflow_run_manager: mount only db + workspace

### LINKED ISSUE #225: workflow_run_manager: mount only a workspace for RJC & RWE
RJC & RWE mount should include only one specific workspace. Current mount `/var/reana/` includes all workspaces.

Mounting only workspace e.g. `/var/reana/users/00000000-0000-0000-0000-000000000000/workflows/09d1fcbe-f5c9-4c32-87c6-513f24c09f0c` was causing following error `psycopg2.OperationalError: FATAL:  could not open relation mapping file "global/pg_filenode.map": Permission denied.
`

roksys:
* Closes #225

### DIFF
--- a/reana_workflow_controller/workflow_run_manager.py
+++ b/reana_workflow_controller/workflow_run_manager.py
@@ -21,6 +21,7 @@ from reana_commons.config import (CVMFS_REPOSITORIES,
                                   WORKFLOW_RUNTIME_USER_UID)
 from reana_commons.k8s.api_client import current_k8s_batchv1_api_client
 from reana_commons.k8s.secrets import REANAUserSecretsStore
+from reana_commons.k8s.volumes import get_shared_volume
 from reana_commons.utils import (create_cvmfs_persistent_volume_claim,
                                  create_cvmfs_storage_class, format_cmd)
 from reana_db.config import SQLALCHEMY_DATABASE_URI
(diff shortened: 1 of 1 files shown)


reana_workflow_controller/workflow_run_manager.py
```diff
@@ -467,3 +457,19 @@ def _create_job_controller_startup_cmd(self, user=None):
             return [full_cmd]
         else:
             return base_cmd.split()
+
+    def _get_volume_mounts(self):
```
diegodelemos: Would [reana_commons.k8s.volumes:get_shared_volume](https://github.com/reanahub/reana-commons/blob/2a5bb3d85d71c12a93541357f10bb7e46bdd6dbe/reana_commons/k8s/volumes.py#L64) be of help here? if not we could adapt it, note that there we are changing depending of the deployment type CEPH or local Minikube, something we miss here.

reana_workflow_controller/workflow_run_manager.py
```diff
@@ -467,3 +457,19 @@ def _create_job_controller_startup_cmd(self, user=None):
...
+        """Return volume mounts."""
+        worksapce = self.workflow.get_workspace()
+        worksapce_mount = {
+            'name': 'default-shared-volume',
+            'mountPath':
+                SHARED_VOLUME_PATH + '/' + worksapce,
+            'subPath': worksapce
+        }
+        db_mount = {
```
diegodelemos: Do we really need the DB mount? can you recheck it? :)

PR #237, opened 2019-08-08
api: create workflow from gitlab project

leticiawanderley:
Signed-off-by: Leticia Farias Wanderley <leticia.farias.wanderley@cern.ch>

### DIFF
--- a/reana_workflow_controller/config.py
+++ b/reana_workflow_controller/config.py
@@ -114,3 +114,12 @@ JOB_CONTROLLER_NAME = 'job-controller'
 
 WORKFLOW_ENGINE_NAME = 'workflow-engine'
 """Default workflow engine container name."""
+
+REANA_GITLAB_HOST = os.getenv('REANA_GITLAB_HOST', 'CHANGE_ME')
+"""GitLab API HOST"""
+
+REANA_GITLAB_URL = 'https://{}'.format(REANA_GITLAB_HOST)
+"""GitLab API URL"""
+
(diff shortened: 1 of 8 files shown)


reana_workflow_controller/config.py
```diff
@@ -114,3 +114,9 @@
 
 WORKFLOW_ENGINE_NAME = 'workflow-engine'
 """Default workflow engine container name."""
+
+GITLAB_URL = os.getenv('GITLAB_URL', 'CHANGE_ME')
```
diegodelemos: Same as https://github.com/reanahub/reana-server/pull/165/files#r313851017

reana_workflow_controller/consumer.py
```diff
@@ -67,12 +70,37 @@ def _update_workflow_status(workflow_uuid, status, logs):
     """Update workflow status in DB."""
     Workflow.update_workflow_status(Session, workflow_uuid,
```
diegodelemos: **Create issue after merge, not to amend on this PR:**

We could have a list of objects who subscribe to the workflow status and all of them implement `update_workflow_status`. Then we just simply iterate and call the method on each of them, so we simplify the implementation for GitLab which we have below.
```python
# Workflow is the DB model, so that subscribes the database to changes
subscribe_to_workflow_status_change = [Workflow, GitLab, Email]
for subscriber in subscribe_to_workflow_status_change:
    subscriber.update_workflow_status(workflow, status)
```

PR #239, opened 2019-08-12
installation: remove celery

diegodelemos:
* Remove Celery code since it won't be installed anymore by REANA-Commons
  (closes reanahub/reana#188).


setup.py
```diff
@@ -44,7 +44,6 @@
 ]
 
 install_requires = [
-    'celery>=4.1.0,<4.3',
```
tiborsimko: Please amend also:

```
tests/test_views.py:477:        # replace celery task with Mock()
tests/test_views.py:492:                # assert the celery task was called with the following
tests/test_views.py:516:        # replace celery task with Mock()
tests/test_views.py:781:        # replace celery task with Mock()
```

PR #245, opened 2019-08-23
setup: reana-commons upgrade

roksys:


diegodelemos:
👍Can you bump REANA-DB as well (https://github.com/reanahub/reana-db/pull/50)

PR #257, opened 2019-10-11
api: stream file uploads using request.stream

diegodelemos:
* After the findings in
  https://github.com/diegodelemos/client-server-flask-big-file-uploads
  it is clear we have to move from multipart uploads to stream uploads
  using `request.stream` directly (closes reanahub/reana-client#302).


reana_workflow_controller/rest.py
```diff
@@ -456,7 +466,7 @@ def upload_file(workflow_id_or_name):
                                    'Please set your REANA_WORKON environment'
                                    'variable appropriately.'.
                                    format(workflow_id_or_name)}), 404
-    except (KeyError, ValueError) as e:
+    except (KeyError, ) as e:
```
tiborsimko: can also remove tuple if we keep one exception only

diegodelemos:
Updated.

PR #270, opened 2019-12-11
k8s: use apps and networking APIs

### LINKED ISSUE #267: interactive sessions: fix after recent k8s/helm update
After recent k8s/helm update, the interactive sessions stopped working:

```
$ reana-client open jupyter -w root6-roofit.serial
Interactive session could not be opened: 
(500)
Reason: Internal Server Error
HTTP response headers: HTTPHeaderDict({'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'Date': 'Fri, 29 Nov 2019 11:57:39 GMT', 'Content-Length': '240'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"no kind \"Ingress\" is registered for version \"extensions/v1beta1\" in scheme \"k8s.io/apimachinery/pkg/apis/meta/internalversion/register.go:31\"","code":500}
```

diegodelemos:
* Extensions API gets will get deprecated Ingressses this commit
  switches to the new Networking API.

* Change deletion of Deployments to use `apps` API which was missing
  in the Kubernetes 1.16 upgrade (closes #267).

### DIFF
--- a/reana_workflow_controller/k8s.py
+++ b/reana_workflow_controller/k8s.py
@@ -13,7 +13,7 @@ from kubernetes.client.rest import ApiException
 from reana_commons.config import REANA_WORKFLOW_UMASK
 from reana_commons.k8s.api_client import (current_k8s_appsv1_api_client,
                                           current_k8s_corev1_api_client,
-                                          current_k8s_extensions_v1beta1)
+                                          current_k8s_networking_v1beta1)
 from reana_commons.k8s.volumes import get_shared_volume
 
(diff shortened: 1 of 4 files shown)


diegodelemos:
This is failing because it needs R-Commons release. Tests are passing locally with the correct version, so it should be fine once released.

roksys:
Commit message might need some rephrasing.

> Extensions API gets will get deprecated Ingressses this commit switches to the new Networking API.

PR #271, opened 2019-12-12
workflow_run_manager: changes workspace group

roksys:
* Addresses reanahub/reana-workflow-controller/issues/269

### DIFF
--- a/reana_workflow_controller/workflow_run_manager.py
+++ b/reana_workflow_controller/workflow_run_manager.py
@@ -441,8 +441,9 @@ class KubernetesWorkflowRunManager(WorkflowRunManager):
                 WORKFLOW_RUNTIME_USER_UID,
                 WORKFLOW_RUNTIME_USER_GID,
                 user)
-            chown_workspace_cmd = 'chown -R {} {};'.format(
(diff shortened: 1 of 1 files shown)


reana_workflow_controller/workflow_run_manager.py
```diff
@@ -441,8 +441,9 @@ def _create_job_controller_startup_cmd(self, user=None):
                 WORKFLOW_RUNTIME_USER_UID,
                 WORKFLOW_RUNTIME_USER_GID,
                 user)
-            chown_workspace_cmd = 'chown -R {} {};'.format(
+            chown_workspace_cmd = 'chown -R {}:{} {};'.format(
                 WORKFLOW_RUNTIME_USER_UID,
+                WORKFLOW_RUNTIME_USER_GID,
```
tiborsimko: If we have `WORKFLOW_RUNTIME_USER_UID` set to 1000 by default, and if a user specifies `kubernetes_uid=500` in the workflow definition, will the workspace be writable?  Thanks to GID=0 it might, but I wonder whether the kubernetes_uid and (future) kubernetes_gid should be consumed already here in R-W-C if we do any kind of chmod/chown stuff...

PR #274, opened 2019-12-20
release: v0.6.0

roksys:


CHANGES.rst
```diff
@@ -1,5 +1,16 @@
 Changes
 =======
+
+Version 0.6.0 (2019-12-20)
+--------------------------
+
+- Moves workflow engine and job controller to the same pod (sidecar pattern)
```
diegodelemos: As I read it now, it sounds like I was reading R-Cluster changes, what do you think of something like this:
```
- Modifies the batch workflow run creation, including an instance of REANA-Job-Controller running
  side by side with the workflow engine (sidecar pattern). Only DB and workflow space are mounted. 
```

CHANGES.rst
```diff
@@ -1,5 +1,16 @@
...
+  and mounts only DB and workflow workspace.
+- Provides user secrets to the job controller.
+- Extends workflow APIs for GitLab integration.
+- Allows stream file uploads.
```
diegodelemos: Thanks for this 🙏 , I've forgotten it in RS release PR!

PR #276, opened 2020-01-07
workflow_run_manager: does not return closed sessions to cli

roksys:
* Closes reanahub/reana-client/issues/310


reana_workflow_controller/rest.py
```diff
@@ -183,7 +183,9 @@ def get_workflows():  # noqa
                                  strftime(WORKFLOW_TIME_FORMAT),
                                  'size': '-'}
             if type == 'interactive':
-                if not workflow.interactive_session_type:
+                if not workflow.interactive_session_type or \
```
tiborsimko: `_type` is present twice

PR #280, opened 2020-01-13
workflow_run_manager: add code mount in dev mode

diegodelemos:
* Adds REANA code mounted to runtime pods (RWE_ and RJC) when
  FLASK_ENV is `development`. We assume that when REANA is installed
  in dev mode, minikube will have all REANA source code under `/code`
  this safe to assume since it is part of our Makefile procedures
  (we do not allow to run REANA in dev mode without the mount active).

### DIFF
--- a/reana_workflow_controller/workflow_run_manager.py
+++ b/reana_workflow_controller/workflow_run_manager.py
@@ -329,7 +329,7 @@ class KubernetesWorkflowRunManager(WorkflowRunManager):
             template=client.V1PodTemplateSpec())
         spec.template.metadata = workflow_metadata
 
-        workflow_enginge_container = client.V1Container(
+        workflow_engine_container = client.V1Container(
             name=current_app.config['WORKFLOW_ENGINE_NAME'],
             image=image,
             image_pull_policy='IfNotPresent',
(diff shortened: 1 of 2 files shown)


reana_workflow_controller/workflow_run_manager.py
```diff
@@ -425,6 +421,28 @@ def _create_job_spec(self, name, command=None, image=None,
...
+            )
+            k8s_code_volume.host_path = client.V1HostPathVolumeSource(
+                code_mount_path
+            )
+            spec.template.spec.volumes.append(k8s_code_volume)
+
+            for container in spec.template.spec.containers:
+                container.env.extend(current_app.config['DEBUG_ENV_VARS'])
+                sub_path = f'reana-{container.name}'
+                if container.name == 'workflow-engine':
```
diegodelemos: I am not very happy about having to put this `if` here... the other option I see is to change the name of the workflow engine containers from `workflow-engine` to `workflow-engine-{self.workflow.type_}`. The biggest disadvantage I see with that approach is that retrieving logs for the workflow engines will differ from each implementation, which if I think about it, it will annoy me more than thinking that there is an `if` here ☺️:
```diff
-$ kubectl logs batch-yadage-d1dbdb66-fbe9-41d9-9509-1fb7ef947f0f-pdph5 workflow-engine
+$ kubectl logs batch-yadage-d1dbdb66-fbe9-41d9-9509-1fb7ef947f0f-pdph5 workflow-engine-yadage
-$ kubectl logs batch-cwl-d1dbdb66-fbe9-41d9-9509-1fb7ef947f0f-pdph5 workflow-engine
+$ kubectl logs batch-cwl-d1dbdb66-fbe9-41d9-9509-1fb7ef947f0f-pdph5 workflow-engine-cwl
-$ kubectl logs batch-serial-d1dbdb66-fbe9-41d9-9509-1fb7ef947f0f-pdph5 workflow-engine
+$ kubectl logs batch-serial-d1dbdb66-fbe9-41d9-9509-1fb7ef947f0f-pdph5 workflow-engine-serial
```

PR #283, opened 2020-01-16
workflow_run_manager: pass workflow file if exists

diegodelemos:
* There will be cases (i.e. integration with Git providers) in
  which there won't be a workflow specification stored in the
  database. This only happens when a submission doesn't come from a REANA
  client. So in these cases, by knowing the path of the workflow
  file, the workflow engines can load with their specific libraries
  the workflow files (which we will optimistically assume are there
  if not, workflow engine should fail saying there is no workflow
  to be processed) (closes reanahub/reana#235).

### DIFF
--- a/reana_workflow_controller/workflow_run_manager.py
+++ b/reana_workflow_controller/workflow_run_manager.py
@@ -63,6 +63,7 @@ class WorkflowRunManager():
                             "--workflow-uuid {id} "
                             "--workflow-workspace {workspace} "
                             "--workflow-json '{workflow_json}' "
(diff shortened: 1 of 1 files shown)


roksys:
Commit message

> There will be cases (i.e. integration with Git providers) in
> which there won't be a workflow specification stored in the
> database. This only happens when a submission comes from **a REANA**
> **client**

Shouldn't it be ?
> This only happens when a submission comes **not from a REANA **client**

diegodelemos:
Updated.

PR #285, opened 2020-01-20
workflow_run: notify about third party failures

diegodelemos:
Closes reanahub/reana#253.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -958,6 +958,14 @@
                 "message": "Status resume is not supported yet."
               }
             }
+          },
+          "502": {
+            "description": "Request failed. Connection to a third party system has failed.",
+            "examples": {
+              "application/json": {
+                "message": "Connection to database timed out, please retry."
+              }
+            }
           }
         },
         "summary": "Set workflow status."
--- a/reana_workflow_controller/errors.py
+++ b/reana_workflow_controller/errors.py
@@ -27,3 +27,7 @@ class REANAWorkflowDeletionError(Exception):
 
 class REANAInteractiveSessionError(Exception):
     """Error when trying to create an interactive session."""
+
+
+class REANAExternalCallError(Exception):
+    """Error when connecting to an external service."""
--- a/reana_workflow_controller/rest/utils.py
+++ b/reana_workflow_controller/rest/utils.py
@@ -25,10 +25,13 @@ from reana_commons.k8s.secrets import REANAUserSecretsStore
 from reana_commons.utils import get_workflow_status_change_verb
 from reana_db.database import Session
 from reana_db.models import Job, JobCache, Workflow, WorkflowStatus
+from sqlalchemy.exc import SQLAlchemyError
+from kubernetes.client.rest import ApiException
 
 from reana_workflow_controller.config import (REANA_GITLAB_HOST,
                                               WORKFLOW_TIME_FORMAT)
-from reana_workflow_controller.errors import (REANAWorkflowControllerError,
+from reana_workflow_controller.errors import (REANAExternalCallError,
+                                              REANAWorkflowControllerError,
                                               REANAWorkflowDeletionError)
 from reana_workflow_controller.workflow_run_manager import \
     KubernetesWorkflowRunManager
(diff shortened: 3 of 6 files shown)


reana_workflow_controller/rest/utils.py
```diff
@@ -25,38 +25,66 @@
...
-        current_db_sessions = Session.object_session(workflow)
         current_db_sessions.add(workflow)
         current_db_sessions.commit()
-        kwrm = KubernetesWorkflowRunManager(workflow)
-        kwrm.start_batch_workflow_run()
-    else:
+
+    def _rollback_start_workflow_db(workflow):
+        workflow.run_started_at = None
+        workflow.status = WorkflowStatus.queued
```
roksys: Just wondering if it shouldn't be ``WorkflowStatus.created``

reana_workflow_controller/rest/utils.py
```diff
@@ -25,38 +25,66 @@
...
             workflow.operational_options = \
                 parameters.get('operational_options')
-        current_db_sessions = Session.object_session(workflow)
         current_db_sessions.add(workflow)
         current_db_sessions.commit()
-        kwrm = KubernetesWorkflowRunManager(workflow)
-        kwrm.start_batch_workflow_run()
-    else:
+
+    def _rollback_start_workflow_db(workflow):
```
roksys: Would it be worth to add `retry` decorator as in https://github.com/reanahub/reana-job-controller/blob/2688996c44c2a1b275076dc58990ad259cd9db7a/reana_job_controller/htcondorcern_job_manager.py#L191? Since this is really crucial function.

reana_workflow_controller/errors.py
```diff
@@ -27,3 +27,7 @@ class REANAWorkflowDeletionError(Exception):
 
 class REANAInteractiveSessionError(Exception):
     """Error when trying to create an interactive session."""
+
+
+class REANAExternalCallError(Exception):
+    """Error when trying to create an interactive session."""
```
mvidalgarcia: Docstring doesn't correspond, probably due to copy paste.

docs/openapi.json
```diff
@@ -1231,4 +1239,4 @@
   },
   "swagger": "2.0",
   "tags": []
-}
+}
```
mvidalgarcia: endline missing

reana_workflow_controller/rest/utils.py
```diff
@@ -25,38 +25,66 @@
...
     KubernetesWorkflowRunManager
 
 
 def start_workflow(workflow, parameters):
     """Start a workflow."""
-    if workflow.status in [WorkflowStatus.created, WorkflowStatus.queued]:
+    current_db_sessions = Session.object_session(workflow)
+    kwrm = KubernetesWorkflowRunManager(workflow)
+
+    def _start_workflow_db(workflow, parameters):
```
mvidalgarcia: For the sake of readability, what do you think about moving these private functions to the top of `start_workflow`?

reana_workflow_controller/rest/utils.py
diegodelemos: Solved IRL :)

reana_workflow_controller/rest/utils.py
diegodelemos: Right now there would be a retry happening the following way:
```console
RS                     RWC                    DBoD       MQ
|                       |                       |         | Q = [X]
|<----------------------|-[RS consumes wf X]----|---------| Q = []
|--[start workflow X]-->|                       |         |
|                       | --[commit to DB]--->  |         |
|                       | --* Fails to          |         | 
|                       |   | connect to DB     |         |
|                       |   | (or other failure)|         |
|                       | <-*                   |         |
|<--------[502]---------|                       |         |
|-----------------------|-[requeues workflow X]-|-------->| Q = [X]
|<----------------------|-[RS consumes wf X]----|---------| Q = []
|---[start workflow X]->|                       |
|                       |----[commit to DB]---->|
|                       |<---[success]----------|
|<--------[200]---------|
```

It is true that if we do it here, in RWC code, we avoid all this going back and forth... But this requeueing mechanism we need for any other failure, so it can serve the two purposes, WDYT? We could ticketise the `retry` for later since no other service will call RWC except RS.

PR #288, opened 2020-01-27
global: changes get_workspace method with workspace_path field

roksys:
* It allows reuse of workspace.

* Addresses reanahub/reana-workflow-controller/issues/289


reana_workflow_controller/workflow_run_manager.py
```diff
@@ -179,7 +179,7 @@ def _workflow_engine_command(self, overwrite_input_parameters=None,
         return (WorkflowRunManager.engine_mapping[self.workflow.type_]
                 ['command'].format(
                     id=self.workflow.id_,
-                    workspace=self.workflow.get_workspace(),
+                    workspace=self.workflow.workspace_path,
```
tiborsimko: We can now upgrade the minimal version of `reana-db` to `0.7.0.dev20200205` in `setup.py`.

PR #295, opened 2020-02-06
rest: load json boolean

diegodelemos:
* `true` and `false` when passed to `bool()` are always true
  (closes reanahub/reana-client#367).


reana_workflow_controller/rest/workflows.py
```diff
@@ -155,7 +155,7 @@ def get_workflows():  # noqa
         user_uuid = request.args['user']
         user = User.query.filter(User.id_ == user_uuid).first()
         type = request.args.get('type', 'batch')
-        verbose = request.args.get('verbose', False)
+        verbose = json.loads(request.args.get('verbose', 'false').lower())
```
tiborsimko: Do we have the same issue with the following?

```
reana_workflow_controller/rest/workflows.py:554:        brief = request.args.get('brief', False)
```

PR #296, opened 2020-02-07
rest: allows to delete queued workflows

roksys:
* Closes reanahub/reana-client/issues/239

### DIFF
--- a/reana_workflow_controller/rest/utils.py
+++ b/reana_workflow_controller/rest/utils.py
@@ -32,7 +32,8 @@ from reana_workflow_controller.config import (REANA_GITLAB_HOST,
                                               WORKFLOW_TIME_FORMAT)
 from reana_workflow_controller.errors import (REANAExternalCallError,
                                               REANAWorkflowControllerError,
-                                              REANAWorkflowDeletionError)
+                                              REANAWorkflowDeletionError,
+                                              REANAWorkflowStatusError)
 from reana_workflow_controller.workflow_run_manager import \
     KubernetesWorkflowRunManager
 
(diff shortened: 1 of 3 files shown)


reana_workflow_controller/rest/utils.py
```diff
@@ -180,27 +180,32 @@ def delete_workflow(workflow,
...
-            to_be_deleted += Session.query(Workflow).\
-                filter(Workflow.name == workflow.name,
-                       Workflow.status != WorkflowStatus.running).all()
-        for workflow in to_be_deleted:
-            if hard_delete:
-                remove_workflow_workspace(workflow.workspace_path)
-                _delete_workflow_row_from_db(workflow)
-            else:
-                if workspace:
+                           WorkflowStatus.failed,
```
diegodelemos: Could you confirm what is happening with the [`reana-server` `scheduler` container](https://github.com/reanahub/reana-server/blob/fb00c8b049b0871a63fc46d9bddae9dc280052ad/reana_server/scheduler.py#L79)? Because we have this operation to delete from the DB, but the workflow will still be present in the RabbitMQ queue... (which one can check with `rabbitmqctl list_queues` inside the `message-broker` component).

I am afraid that `reana-server` will continuously try to start the workflow, fail, requeue, try to start, fail, requeue... infinitely.

If that is the case, I feel we need to find a nice way of doing this operation atomically.

roksys: This [PR](https://github.com/reanahub/reana-server/pull/214) is solving this.

diegodelemos:
Can you rebase? I will test it :)

PR #297, opened 2020-02-10
rest: adds a possibility to query logs for a given step

roksys:
* Addresses reanahub/reana-client/issues/369

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -504,6 +504,20 @@
             "name": "workflow_id_or_name",
             "required": true,
             "type": "string"
+          },
+          {
+            "description": "Steps of a workflow.",
+            "in": "body",
+            "name": "steps",
+            "required": false,
+            "schema": {
+              "description": "List of step names to get logs for.",
+              "items": {
+                "description": "Step name.",
+                "type": "string"
+              },
+              "type": "array"
+            }
           }
         ],
         "produces": [
--- a/reana_workflow_controller/rest/utils.py
+++ b/reana_workflow_controller/rest/utils.py
@@ -21,6 +21,7 @@ from reana_commons.config import REANA_WORKFLOW_UMASK
 from reana_commons.k8s.secrets import REANAUserSecretsStore
 from reana_commons.utils import get_workflow_status_change_verb
 from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy import and_
 
 import fs
 from flask import current_app as app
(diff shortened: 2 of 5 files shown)


docs/openapi.json
```diff
@@ -504,6 +504,20 @@
...
             "required": true,
             "type": "string"
+          },
+          {
+            "description": "Steps of a workflow.",
+            "in": "body",
+            "name": "steps",
+            "required": false,
+            "schema": {
+              "description": "List of steps names to get logs.",
```
tiborsimko: "List of step names to get logs for."

docs/openapi.json
```diff
@@ -504,6 +504,20 @@
...
+              "items": {
+                "description": "step name.",
```
tiborsimko: If we end with a dot, then let's make it a phrase, as it were, so starting with capital: "Step name."

reana_workflow_controller/rest/utils.py
```diff
@@ -120,10 +121,16 @@ def get_workflow_name(workflow):
     return workflow.name + '.' + str(workflow.run_number)
 
 
-def build_workflow_logs(workflow):
+def build_workflow_logs(workflow, steps=None):
```
diegodelemos: Not to do now, nothing to do with this PR. If the answer is yes for any of the following points, we could create an issue and do it later, if not just ignore this comment:

<details><summary>Does `build_workflow_logs` function name sound confusing? (it returns step logs, to retrieve workflow engine logs we do `workflow.logs`</summary>
<p>
- Rename this function: currently, it sounds like it is returning the workflow logs... which in any case doesn't sound very clear to me, for example, `build_workflow_step_logs` or `build_workflow_job_logs` would be more explicit. Moreover, this function could be moved to the REANA-DB models.
</p>
</details>


<details><summary>Does it sound confusing that in the code base we have `steps` and `jobs` which in most of the cases is the same?</summary>
<p>
- We should clarify when we do the translation of `step` (user-centred) to `job` (backend jargon) ... or if we should do it at all (there is a lot of history, jobs are called jobs in the backend because of the compute backends, steps are called steps because it is how they are normally referred in the workflow world), just because code wise this might sound confusing?
</p>
</details>

docs/openapi.json
```diff
@@ -1242,4 +1256,4 @@
   },
   "swagger": "2.0",
   "tags": []
-}
+}
```
mvidalgarcia: Missing end-line. Perhaps we can amend `generate_openapi_spec.py` to avoid this? It happens to me all the time as well.

reana_workflow_controller/rest/utils.py
mvidalgarcia: > Does `build_workflow_logs` function name sound confusing? (it returns step logs, to retrieve workflow engine logs we do `workflow.logs`

It's indeed confusing, I thought they were returning kind of the same. We should point out somehow the difference between **workflow** logs and **job** logs (if I understand it correctly).

> Does it sound confusing that in the code base we have `steps` and `jobs` which in **most of the cases** is the same?

What do you mean by _in most of the cases_? If they are actually the same, then yes, I find it confusing as well.

diegodelemos: For example, in serial workflows, when you specify this:
```yaml
....
workflow:
  type: serial
  specification:
    steps:
      - name: gendata ## This is a workflow step
        environment: 'reanahub/reana-env-root6'
        commands:
        - mkdir -p results ## This is a compute job
        - root -b -q 'code/gendata.C(${events},"${data}") ## This is a compute job
....
```

roksys: Not always the same, one step can have multiple jobs. It's the case for serial roofit example, `gendata` step has two jobs.

docs/openapi.json
diegodelemos: This change is missing.

diegodelemos:
You can bump R-Commons to `0.7.0.dev20200211` :)

PR #301, opened 2020-02-13
workflow-run-manager: mount CVMFS on int. sessions

### LINKED ISSUE #300: workflow-run-manager: interactive sessions do not have access to CVMFS
REANA batch workflows at CERN have access to CVMFS, but this volume is no available on interactive sessions:
```
$ # on a terminal inside an interactive session
$ ls /cvmfs/
ls: /cvmfs: No such file or directory
```
We should be consistent.

diegodelemos:
* Closes #300.

### DIFF
--- a/reana_workflow_controller/k8s.py
+++ b/reana_workflow_controller/k8s.py
@@ -10,11 +10,11 @@ import os
 
 from kubernetes import client
 from kubernetes.client.rest import ApiException
-from reana_commons.config import REANA_WORKFLOW_UMASK
+from reana_commons.config import CVMFS_REPOSITORIES, REANA_WORKFLOW_UMASK
 from reana_commons.k8s.api_client import (current_k8s_appsv1_api_client,
                                           current_k8s_corev1_api_client,
                                           current_k8s_networking_v1beta1)
-from reana_commons.k8s.volumes import get_shared_volume
+from reana_commons.k8s.volumes import get_k8s_cvmfs_volume, get_shared_volume
 
 from reana_workflow_controller.config import (  # isort:skip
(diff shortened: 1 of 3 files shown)


reana_workflow_controller/k8s.py
```diff
@@ -145,6 +148,43 @@ def add_reana_shared_storage(self):
...
+        for repo_name, path in cvmfs_map.items():
+            volume = get_k8s_cvmfs_volume(repo_name)
+            volumes.append(volume)
+            volume_mounts.append(
+                {'name': volume['name'],
+                 'mountPath': '/cvmfs/{}'.format(path),
+                 'readOnly': volume['readOnly']}
+            )
+
+        return volumes, volume_mounts
```
diegodelemos: This is the piece of code that should be moved to R-Commons, and [reused in RJC](https://github.com/reanahub/reana-job-controller/blob/fb82d77063a3051153f7868c4a2de5a6680893a4/reana_job_controller/kubernetes_job_manager.py#L148-L165).

setup.py
```diff
@@ -46,7 +46,6 @@
 install_requires = [
     'Flask-SQLAlchemy>=2.2',
     'Flask>=0.12',
-    'Flask-Login<0.5.0',  # FIXME waiting for Flask-Security release
```
tiborsimko: Is removing `Flask-Login` pin safe here? (no invenio in r-w-controller)

diegodelemos: It was my bad, I shouldn't have pinned in the first place since there are no `Flask-Login` nor `Flask-Security` in RWC 😅.

PR #306, opened 2020-02-28
config: load configured image pull secrets

diegodelemos:
* Closes reanahub/reana-job-controller#230.

### DIFF
--- a/reana_workflow_controller/config.py
+++ b/reana_workflow_controller/config.py
@@ -120,3 +120,6 @@ REANA_GITLAB_URL = 'https://{}'.format(REANA_GITLAB_HOST)
 
 REANA_URL = os.getenv('REANA_URL', 'CHANGE_ME')
 """REANA URL"""
+
+IMAGE_PULL_SECRETS = os.getenv('IMAGE_PULL_SECRETS', '').split(',')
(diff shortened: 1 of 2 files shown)


danikam:
Hi @tiborsimko, @diegodelemos, thanks for this! Just to make sure I understand - would the env var `IMAGE_PULL_SECRETS` be the gitlab access token?

diegodelemos:
Hello @danikam, in principle, you just need to add [REANA@gitlab.cern.ch](https://gitlab.cern.ch/reana) as Reporter member to the project where you host the images and then reana.cern.ch will be able to pull images from there, see:
Screenshot 2020-02-28 at 14 56 11

danikam:
Ah, I see - thanks @diegodelemos! So the user shouldn't need to set this `IMAGE_PULL_SECRETS`  env var?

diegodelemos:
No :) as long as you add the REANA user as a collaborator in the GitLab project there is nothing else you need to do, it is transparent (this `IMAGE_PULL_SECRETS ` is an internal environment variable, only the REANA system can access it).

danikam:
Got it, thanks! Will start to test this out asap with the monojet repo images :)

Issue #149, opened 2018-11-22
state: use state machine to manage workflow state


diegodelemos:
As mentioned in https://github.com/reanahub/reana-client/issues/202#issuecomment-441049790, we will need yet another state for workflows. This idea has started on https://github.com/reanahub/reana-client/issues/192#issuecomment-437003151 where we defined the first state machine which represent the actions that can be taken given the status of a workflow.

Right now we have this logic spreaded all over the code, i.e. for [`starting`](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/rest.py#L1390) or [`deleting`](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/rest.py#L1680-L1689).

We should centralise this logic so we can easily change and maintain how state is managed, some alternatives are:
- Implementing ourselves the logic for the state machine in an isolated module.
- Use an existing solution for state machines, i.e. [`pytransitions`](https://github.com/pytransitions/transitions).

diegodelemos:
While tackling https://github.com/reanahub/reana-client/issues/395 we have come up with a new method for the workflow model, [`Workflow.can_transition_to()`](https://github.com/reanahub/reana-db/blob/89f6d0ca4d13aabf16e86344969fded07d0c088a/reana_db/models.py#L264). We should replace all the decentralised logic ([`workflow_controller.utils.stop_workflow`](https://github.com/reanahub/reana-workflow-controller/blob/8ba38e9b44621f58c63da1fd4403f5b4065910fb/reana_workflow_controller/rest/utils.py#L101), [`workflow_controller.utils.delete_workflow`](https://github.com/reanahub/reana-workflow-controller/blob/8ba38e9b44621f58c63da1fd4403f5b4065910fb/reana_workflow_controller/rest/utils.py#L189-L194) etc ...) to use the general approach `workflow. can_transition_to(next_status)`.

PR #310, opened 2020-04-07
workflow_run_manager: workflow stop race condition

diegodelemos:
* Happening when user stops a workflow before its first job is created.
  RWC stops only the existing jobs. However, because there is a grace
  period for stopping pods, RJC sidecar still runs, submitting a new
  job, and reporting its status, causing the workflow to “revive”
  (closes reanahub/reana-client#395).

### DIFF
--- a/reana_workflow_controller/config.py
+++ b/reana_workflow_controller/config.py
@@ -11,7 +11,7 @@
 import os
 
 from packaging.version import parse
-from reana_commons.config import SHARED_VOLUME_PATH
+from reana_commons.config import REANA_COMPONENT_PREFIX, SHARED_VOLUME_PATH
 
 from reana_workflow_controller.version import __version__
 
@@ -67,7 +67,8 @@ WORKFLOW_ENGINE_COMMON_ENV_VARS = [
 """Common to all workflow engines environment variables."""
 
 DEBUG_ENV_VARS = ({'name': 'WDB_SOCKET_SERVER',
-                   'value': 'reana-wdb'},
+                   'value': os.getenv('WDB_SOCKET_SERVER',
+                                      f'{REANA_COMPONENT_PREFIX}-wdb')},
                   {'name': 'WDB_NO_BROWSER_AUTO_OPEN',
                    'value': 'True'},
                   {'name': 'FLASK_ENV',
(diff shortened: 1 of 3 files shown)


diegodelemos:
- Quick solution: Force delete with grace period 0
    - Didn't work, still is not 0 seconds to delete -> ❌
- Better solution: Ignore status updates for workflows stopped in DB
    - `stop_workflow` performs an [atomic operation](https://github.com/reanahub/reana-workflow-controller/blob/79a64bcf007a3cca91b2b94c873b2582ac795a71/reana_workflow_controller/rest/utils.py#L99-L112), therefore, if the stop of a workflow and its alive jobs was done, the workflow has successfully been marked as stopped in the database.
    - Use that field to ignore, in [`job-status-consumer`](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/consumer.py), status updates for stopped workflows (something similar is being done with GitLab, see [here](https://github.com/reanahub/reana-workflow-controller/blob/79a64bcf007a3cca91b2b94c873b2582ac795a71/reana_workflow_controller/consumer.py#L99))
- Best solution: Create the state machine for workflow status manipulation and embed logic there, see [previous musings](https://github.com/reanahub/reana-workflow-controller/issues/149).

diegodelemos:
I haven't fully addressed the state machine but I've recorded the knowledge summarised in https://github.com/reanahub/reana-client/issues/192#issuecomment-437003151 (referenced in https://github.com/reanahub/reana-workflow-controller/issues/149) inside a function in a separate module prepared for a possible future refactor for status transition management.

reana_workflow_controller/workflow_state.py
```diff
@@ -0,0 +1,34 @@
...
+# Copyright (C) 2020 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+"""Workflow status transition management."""
+
+from reana_db.models import WorkflowStatus
+
+
+def can_transition(workflow, next_status):
```
mvidalgarcia, line 13: Looks good! What about moving this function to the Workflow model? So we can do `workflow.can_transition_to(next_status)`

diegodelemos: https://github.com/reanahub/reana-db/pull/63

PR #312, opened 2020-04-28
VC3 Integration

CodyKank:
Minor changes / additions for VC3

- Added environment variables in config.py
- Shared /tmp space on with job controller side car in workflow_run_manager.py


setup.py
```diff
@@ -44,13 +44,17 @@
...
+    # and conflicting with requests 2.20.0
+    'urllib3<1.25.0',
     'Flask-SQLAlchemy>=2.2',
     'Flask>=0.12',
     'fs>=2.0',
     'gitpython>=2.1',
     'jsonpickle>=0.9.6',
     'marshmallow>2.13.0,<=2.20.1',
     'packaging>=18.0',
+    'kubernetes==10.0.1',
```
khurtado, line 57: Note we can likely get rid of this changes at this point (same with a few other packages that workaround this issue in the same way). We have the testing from the change in reana-commons pending.

CodyKank: Fixed, the workaround and the comments have been removed.

PR #317, opened 2020-07-08
tests: move to containerised postgres DB

mvidalgarcia:
partially addresses reanahub/reana#348
depends on reanahub/pytest-reana#63 reanahub/reana-db#73

### DIFF
--- a/reana_workflow_controller/config.py
+++ b/reana_workflow_controller/config.py
@@ -10,7 +10,6 @@
 
 import os
 
-from packaging.version import parse
 from reana_commons.config import REANA_COMPONENT_PREFIX, SHARED_VOLUME_PATH
 
 from reana_workflow_controller.version import __version__
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -1,16 +1,73 @@
-#!/bin/sh
+#!/bin/bash
 #
 # This file is part of REANA.
-# Copyright (C) 2017, 2018, 2019 CERN.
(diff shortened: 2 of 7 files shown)


reana_workflow_controller/config.py
```diff
@@ -15,6 +15,8 @@
 
 from reana_workflow_controller.version import __version__
 
+from reana_db.config import SQLALCHEMY_DATABASE_URI
+
```
tiborsimko, line 19: Imported but not used?

run-tests.sh
```diff
@@ -1,16 +1,72 @@
...
+    OLD="$(docker ps --all --quiet --filter=name=reana-postgres)"
+    if [ -n "$OLD" ]; then
+        echo '==> [INFO] Cleaning old DB container...'
+        docker stop reana-postgres
+    fi
+}
+
+start_db_container() {
+    echo '==> [INFO] Starting DB container...'
+    docker run --rm --name reana-postgres -p 5432:5432 -e POSTGRES_PASSWORD=mysecretpassword -d postgres:9.6.2
```
tiborsimko, line 46: `reana-postgres` : it would be nice to use the same container name as in `reana-dev ...` commands... Was it `postgres__reana-workflow-controller`?

reana_workflow_controller/config.py
mvidalgarcia: It's used to pick `REANA_SQLALCHEMY_DATABASE_URI` env var from `run-tests.sh` but I've moved it to conftest.py

PR #319, opened 2020-07-14
rest: pagination

mvidalgarcia:
closes reanahub/reana-server#180


reana_workflow_controller/rest/utils.py
```diff
@@ -538,3 +533,69 @@ def get_workflow_progress(workflow):
...
+                        to_idx=req["page"] * req["size"],
+                        links=dict(
+                            prev={"page": req["page"] - 1},
+                            self={"page": req["page"]},
+                            next={"page": req["page"] + 1},
+                        ),
+                    )
+                )
+
+            def paginate(query_or_list):
```
diegodelemos, line 574: What do you think of adding a docstring to this method explaining that the query it might take is an SQLAlchemy query object? I think we might be confused with this function in the future if we don't explain it well.

mvidalgarcia: Yeah, definitely there's some technical debt in this part. I can add the docstring, no problem. I was also thinking about creating a `Pagination` class and populate instances of it to be returned in this function, instead of using a simple dictionary, what do you think? It can be another issue

diegodelemos: For me, it is fine as it is, as long as it is explained. We can create a separate issue for refactoring it and now is the best time as you have it fresh :)

mvidalgarcia: https://github.com/reanahub/reana-workflow-controller/issues/321

PR #324, opened 2020-07-24
workflow-run-manager: use node selector for RWE's

diegodelemos:
* If there is a configured node label to use as nodeSelector
  RWC will use it so the RWE's pods get scheduled to those nodes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ install_requires = [
     "jsonpickle>=0.9.6",
     "marshmallow>2.13.0,<=2.20.1",
     "packaging>=18.0",
-    "reana-commons[kubernetes]>=0.7.0a1,<0.8.0",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -52,7 +52,7 @@
     "jsonpickle>=0.9.6",
     "marshmallow>2.13.0,<=2.20.1",
     "packaging>=18.0",
-    "reana-commons[kubernetes]>=0.7.0a1,<0.8.0",
+    "reana-commons[kubernetes]>=0.7.0a2,<0.8.0",
```
tiborsimko, line 55: Just a post-merge note: it is actually better if the r-commons version bump was coming as part of the feature commit, not as a new commit, to better express that "this feature neeeds new r-commons".  Could be important for future bisecting. (Saying in general; in this case the relation is immediate and simple as the commits are very adjacent.)

PR #334, opened 2020-09-22
rest: retreive workspace disk usage from db models

audrium:
closes: https://github.com/reanahub/reana-server/issues/279


reana_workflow_controller/rest/workflows.py
```diff
@@ -207,18 +204,13 @@ def get_workflows(paginate=None):  # noqa
...
-                        workflow.workspace_path
-                    )
-                    disk_usage_info = get_workspace_disk_usage(
-                        absolute_workspace_path, block_size=block_size
-                    )
-                    if disk_usage_info:
-                        workflow_response["size"] = disk_usage_info[-1]["size"]
-                    else:
-                        workflow_response["size"] = "0K"
+                disk_usage_info = workflow.get_workspace_disk_usage(
```
diegodelemos, line 207: I know it was done like this before, but, couldn't we use the `summarize` option and get just one element? I think that if we do the simplification stated in https://github.com/reanahub/reana-db/pull/84/files#r492677834 this code should become simpler.

audrium: Yes, indeed. Nice catch!

diegodelemos (approved):

Issue #332, opened 2020-09-21
quotas: calculate quota usage for workflows and store them


diegodelemos:
Fill the new fields/table created in https://github.com/reanahub/reana-db/issues/83.

- CPU: We could use the [existing information about workflow start, stop/finished](https://github.com/reanahub/reana-db/blob/c89d3703b8bcae2f8af83f6c81473b201e1997a0/reana_db/models.py#L255-L257)
    - Note we shouldn't allow deletion of CPU usage
- HDD: RWC has full access to the workflow workspace
    - It should store the workspace disk usage in DB after the workflow finishes
    - The disk usage would be updated every night?

tiborsimko:
See also https://github.com/reanahub/reana-workflow-controller/issues/333#issuecomment-696627891 about updating DB numbers via several triggers

PR #336, opened 2020-09-30
models: REANA-DB to manage workflow status change

diegodelemos:
* Let REANA-DB atomically manage status change related operations.
  For example, setting the finished/stopped/failed time (removed
  from RWC in this PR), updating quota usage etc...

Closes reanahub/reana-workflow-controller#332.

### DIFF
--- a/reana_workflow_controller/consumer.py
+++ b/reana_workflow_controller/consumer.py
@@ -112,7 +112,6 @@ def _update_workflow_status(workflow, status, logs):
             WorkflowStatus.queued,
         ]
         if status not in alive_statuses:
-            workflow.run_finished_at = datetime.now()
             _delete_workflow_engine_pod(workflow)
 
 
--- a/reana_workflow_controller/rest/utils.py
+++ b/reana_workflow_controller/rest/utils.py
@@ -47,7 +47,6 @@ def start_workflow(workflow, parameters):
     """Start a workflow."""
 
     def _start_workflow_db(workflow, parameters):
-        workflow.run_started_at = datetime.now()
         workflow.status = WorkflowStatus.running
         if parameters:
             workflow.input_parameters = parameters.get("input_parameters")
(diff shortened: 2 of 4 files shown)


reana_workflow_controller/consumer.py
```diff
@@ -112,7 +112,6 @@ def _update_workflow_status(workflow, status, logs):
             WorkflowStatus.queued,
         ]
         if status not in alive_statuses:
-            workflow.run_finished_at = datetime.now()
```
mvidalgarcia, line 115: Question: Don't we have other occurrences of `workflow.run_finished_at/run_started_at/run_stopped_at` around? If I understood correctly these assignments will only happen when [the listener on `reana-db` is triggered](https://github.com/reanahub/reana-db/pull/87/files#diff-ce53bbb90bfc9cae2aa070829d3c6148R479), won't they?

Doing a search I could find [an occurrence of `run_started_at`](https://github.com/reanahub/reana-workflow-controller/blob/85376079f10e10111643ac515f5ee7079d21b657/reana_workflow_controller/rest/utils.py#L50), is this OK?

diegodelemos: Like always to the point 🎯 , thanks a lot for spotting:
```console
$ rg workflow.run_stopped_at
reana_workflow_controller/rest/utils.py
110:        workflow.run_stopped_at = datetime.now()
```
 I will amend :)

mvidalgarcia: Or maybe I'm wrong... 🤔  I see that specifically [`run_started_at` is set manually](https://github.com/reanahub/reana-workflow-controller/blob/85376079f10e10111643ac515f5ee7079d21b657/reana_workflow_controller/rest/utils.py#L50) while `run_stopped_at` and `run_finished_at` are set in the `reana-db` listener. So my question is, when [this status change happens](https://github.com/reanahub/reana-workflow-controller/blob/85376079f10e10111643ac515f5ee7079d21b657/reana_workflow_controller/rest/utils.py#L51), won't the `reana-db` listener be triggered so we can set `run_started_at` there instead of in `r-w-c`?

mvidalgarcia (approved):

PR #340, opened 2020-10-16
rest: return interactive sessions on verbose workflows request

audrium:
closes https://github.com/reanahub/reana-ui/issues/77

### DIFF
--- a/reana_workflow_controller/workflow_run_manager.py
+++ b/reana_workflow_controller/workflow_run_manager.py
@@ -359,9 +359,11 @@ class KubernetesWorkflowRunManager(WorkflowRunManager):
             )
         finally:
             if action_completed:
-                int_session.status = RunStatus.stopped
+                # TODO: once multiple sessions will be supported instead of
+                # deleting a session, its status should be changed to "stopped"
+                # int_session.status = RunStatus.stopped
                 current_db_sessions = Session.object_session(self.workflow)
-                current_db_sessions.add(int_session)
+                current_db_sessions.delete(int_session)
                 current_db_sessions.commit()
 
     def stop_batch_workflow_run(self):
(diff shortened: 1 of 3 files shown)


reana_workflow_controller/workflow_run_manager.py
```diff
@@ -359,9 +359,11 @@ def stop_interactive_session(self, interactive_session_id):
             )
         finally:
             if action_completed:
-                int_session.status = RunStatus.stopped
+                # TODO: once multiple sessions will be supported instead of
+                # deleting a session, its status should be changed to "stopped"
+                # int_session.status = RunStatus.stopped
                 current_db_sessions = Session.object_session(self.workflow)
-                current_db_sessions.add(int_session)
+                current_db_sessions.delete(int_session)
```
diegodelemos, line 366: For now, deleting the associated interactive session of a workflow on `reana-client close -w wf1` is ok since each workflow can only have one session. However, once we start accounting CPU and disk usage at interactive session level, this would make us loose quota usage information. This can either be a separate issue or can be something to take into account when we do accounting on interactive sessions.

audrium: Totally agree, that's why I left a [todo comment](https://github.com/reanahub/reana-workflow-controller/pull/340/files#diff-0941dd1fa9e589338659a84e2274936332c7498f29b6727873957b99fbb7b1fcR362) above this line to address it as soon as multiple sessions will be supported. If we change it not to be deleted, with our current implementation it would be possible to open only one interactive session per workflow. We need to detach interactive sessions from the workflow in logic as we did in the database part, but it's not a super easy change

mvidalgarcia (approved):

PR #342, opened 2020-10-19
rest: simplify disk usage format

diegodelemos:
* Shows disk usage format either in raw (bytes) or humand-friendly
  format (closes reanahub/reana-server#296).


mvidalgarcia (approved):

reana_workflow_controller/rest/workflows.py
```diff
@@ -238,17 +248,18 @@ def get_workflows(paginate=None):  # noqa
...
-                        workflow_response["size"] = "0K"
-                except ValueError:
-                    workflow_response["size"] = "0K"
+                workflow_response["size"] = (
+                    workflow.get_quota_usage()
+                    .get("disk", {})
+                    .get("usage", empty_disk_usage)
+                )
+            else:
+                workflow_response["size"] = empty_disk_usage
```
diegodelemos, line 262: This is not pretty, but we [cannot specify response properties to be nullable with OpenAPI2](https://stackoverflow.com/questions/48111459/how-to-define-a-property-that-can-be-string-or-null-in-openapi-swagger)... If you have any idea on how to do it better, please share.

mvidalgarcia: Nothing comes to my mind so far.

PR #345, opened 2020-10-26
rest: get rid of workflow hard deletion

mvidalgarcia:
closes reanahub/reana-client#436


diegodelemos (approved):
Since this PR group will probably be merged today together with https://github.com/reanahub/reana-server/issues/296, it would be great if we get https://github.com/reanahub/reana-commons/pull/225 merged also and have only one release today.

PR #347, opened 2020-10-28
ci: added github actions workflow

audrium:


audrium:
Here is an example of GH actions workflow: https://github.com/audrium/reana-workflow-controller/actions/runs/333676260

mvidalgarcia (approved):
LGTM 👏

PR #349, opened 2020-11-09
api: recalculate quota on file upload/deletion

### LINKED ISSUE #339: files: recalculate quota on file upload/deletion
Recalculate quota as in update both `WorkflowResource.quantity_used` and `UserResource.quota_used` for CPU and disk.

File upload/deletion (`reana-client rm`/`reana-client upload`) -> we can run the full calculation for now but, this can become dangerous:
    - Doing this calculation on every upload could be a very expensive server-side operation, as we [upload each file on a different HTTP request](https://github.com/reanahub/reana-client/blob/577e5dd52f14a75bed32863ccbd419e6c2ebac0d/reana_client/cli/files.py#L277).
    - We know how big those files are when uploaded, so in theory we can just increment/decrement the disk usage by the number of bytes the files occupy at workflow and user level.

_Originally posted by @diegodelemos in https://github.com/reanahub/reana-db/pull/92#discussion_r501686934_

mvidalgarcia:
closes #339

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 Changes
 =======
 
+Version master (UNRELEASED)
+---------------------------
+
+- Adds users quota accounting.
+
 Version 0.7.0 (2020-10-20)
 --------------------------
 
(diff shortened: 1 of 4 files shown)


mvidalgarcia:
Test likely failing due a to change in [`reana-db` functions signature](https://github.com/reanahub/reana-db/pull/112/files#diff-88994a905f01648dbb88c29f34110ec0fa9ad4af7a2d2be3e8038167b8f703b3R239).

diegodelemos (approved):
R-DB released needs CI rerun.

Issue #363, opened 2021-02-23
workflow-run-manager: workflow-engine doesn't start and workflow gets stuck in running state


diegodelemos:
## Issue

A general problem in REANA regarding workflows stuck in running state was fixed in https://github.com/reanahub/reana/issues/478. However, there is still a corner case: there is no way to avoid stuck `running` workflows if `workflow-engine` pod never starts.

## Cause

The root cause is that [we set the status to running in RWC before the actual action takes place](https://github.com/diegodelemos/reana-workflow-controller/blob/956691d7a8ade50054496e4c5b822f77c2b2bcd4/reana_workflow_controller/rest/utils.py#L50:L50), it is not guaranteed for the workflow to start any time soon or start at all (it depends on an external system, Kubernetes). And if the workflow doesn't start, there won't be [status changes](https://github.com/reanahub/reana-workflow-engine-yadage/blob/e20432b8d2c20e8c35fb53e4444f98b9612808cb/reana_workflow_engine_yadage/externalbackend.py#L148-L150).

## Solutions

- A possible solution for this would be to introduce a new status (see [current statuses](https://github.com/reanahub/reana-db/blob/a3adc6866ee3233e1c817825709979eb5b6165a2/reana_db/models.py#L286)), e.g. `creating` (TBD better name). This way the flow would look like follows:
  1. User runs workflow -> `queued` ([set by REANA-Server](https://github.com/reanahub/reana-server/blob/d83c88ab92ac7509bccade79e126fc3cc6e25521/reana_server/rest/workflows.py#L911) just before actually queuing it)
  2. REANA-Workflow receives a request to start the workflow -> sets its status to `creating` and requests Kubernetes to start the workflow engine
  3. Kubernetes starts the `workflow-engine` pod -> the workflow engine itself sets the status to `running` when it starts its execution (adding it to the factory [`create_workflow_engine_command `](https://github.com/reanahub/reana-commons/blob/1d89c9419353c2e643216b16d5a3b39a1c16d9aa/reana_commons/workflow_engine.py#L172) so all workflow engines behave the same)

  This way if a workflow gets stuck in `creating` status, we could [allow deletion for workflows in `creating` status](https://github.com/reanahub/reana-db/blob/a3adc6866ee3233e1c817825709979eb5b6165a2/reana_db/models.py#L298). As for orphan workflow engine pods that could result from this, they could be garbage collected (e.g. for all `deleted` workflows, check if workflow engine exists and clean up).

[...]

PR #371, opened 2021-04-21
rest: set workflow status to pending after starting the workflow

audrium:
closes https://github.com/reanahub/reana-workflow-controller/issues/363

Depends on https://github.com/reanahub/reana-db/pull/123 (tests fail because `pending` status is not present yet)

#### Changes

* Sets workflow status to `pending` right after the workflow has been started
* ~~Adds `get_workflow_status_name()` function to return back the workflow status~~
  - ~~`pending` status should be transparent to the user, therefore in that case the `queued` status is returned~~

* ~~Modifies `get_workflows()` endpoint to include `pending` workflows if filtered by `queued` status~~

#### How to test

I've tested it by running simple `helloworld` demo and modifying [reana_workflow_engine_serial](https://github.com/reanahub/reana-workflow-engine-serial/blob/b9219f631f327b6e1f45e66dbc1f6c364ce0e973/reana_workflow_engine_serial/tasks.py#L152) by adding 
```
import time
time.sleep(20)
```
just before ` publish_workflow_start()`. In this case we can simulate that workflow is stuck a bit in `pending` state.

~~In addition [these lines](https://github.com/reanahub/reana-workflow-controller/blob/0f248dc57812f88e76a56e24cc782dd460545cb0/reana_workflow_controller/rest/utils.py#L137-L138) could be commented to inspect the `pending` status using cli (note that UI will break, since the new state is not supported there)~~

### DIFF
--- a/reana_workflow_controller/rest/utils.py
+++ b/reana_workflow_controller/rest/utils.py
@@ -48,7 +48,7 @@ def start_workflow(workflow, parameters):
     """Start a workflow."""
 
     def _start_workflow_db(workflow, parameters):
-        workflow.status = RunStatus.running
+        workflow.status = RunStatus.pending
         if parameters:
             workflow.input_parameters = parameters.get("input_parameters")
             workflow.operational_options = parameters.get("operational_options")
@@ -71,6 +71,7 @@ def start_workflow(workflow, parameters):
                 RunStatus.failed,
                 RunStatus.finished,
                 RunStatus.queued,
+                RunStatus.pending,
             ]:
                 raise REANAWorkflowControllerError(failure_message)
     elif workflow.status not in [RunStatus.created, RunStatus.queued]:
@@ -193,6 +194,7 @@ def delete_workflow(workflow, all_runs=False, workspace=False):
         RunStatus.deleted,
         RunStatus.failed,
         RunStatus.queued,
+        RunStatus.pending,
     ]:
         try:
             to_be_deleted = [workflow]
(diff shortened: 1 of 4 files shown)


reana_workflow_controller/rest/utils.py
```diff
@@ -125,6 +127,18 @@ def get_workflow_name(workflow):
     return workflow.name + "." + str(workflow.run_number)
 
 
+def get_workflow_status_name(workflow):
```
diegodelemos, line 130: Have you tried to move such logic into a new property in REANA-DB? (asking just in case you already found a blocker 😄 ). We could have something like `Workflow.status_ui` (this name is bad, this is just a placeholder) which could be a [hybrid property](https://docs.sqlalchemy.org/en/14/orm/extensions/hybrid.html). This way we could perhaps overwrite the `.in_` method for that property (useful for [filtering](https://github.com/reanahub/reana-workflow-controller/blob/24c3a6222c09bb282ac0153e930077be292eee0f/reana_workflow_controller/rest/workflows.py#L228)), and also we wouldn't break the UI nor the client ([this](https://github.com/reanahub/reana-workflow-controller/blob/24c3a6222c09bb282ac0153e930077be292eee0f/reana_workflow_controller/rest/workflows.py#L237) would access hypothetical `Workflow.status_ui`).

On the one hand, if we do such a split I feel it should be close to the models, on the other hand, I see it might be confusing to decide which one to use. Let me know what you think.

audrium: Created issue to follow up on this [here](https://github.com/reanahub/reana-workflow-controller/issues/375)

diegodelemos (approved):

Issue #204, opened 2019-04-01
refactor: use workflow status as enum instead of integers


diegodelemos:
Currently we are sending the workflow status changes from the workflow engines as the [integer which represents the status](https://github.com/reanahub/reana-db/blob/master/reana_db/models.py#L56-L65), see for example in [RWE Serial](https://github.com/reanahub/reana-workflow-engine-serial/blob/master/reana_workflow_engine_serial/utils.py#L222). This is happening because the [enum](https://github.com/reanahub/reana-db/blob/master/reana_db/models.py#L56-L65) which contains this statuses is inside REANA-DB and the workflow engines do not have DB as a dependency.

Even though it works it could easily lead to errors and it is definitely less readable. Also it would be more clear when dealing with the [state-machine](https://github.com/reanahub/reana-workflow-controller/issues/149) for workflow statuses.

A **possible solution** would be to move these enums (for Workflows and for Jobs) to REANA-Commons.

diegodelemos:
It would be great if we also change status enum real values from integer to strings. E.g.:
```diff
class WorkflowStatus(enum.Enum):
    """Enumeration of possible workflow statuses."""

-    created = 0
+    created = 'created'
...
```


This would:
- Make code more readable and avoid cases like https://github.com/reanahub/reana-workflow-controller/pull/374#discussion_r621082100
- Make database queries/rows more user friendly/readable

PR #374, opened 2021-04-26
rest: add job start and finish times

diegodelemos:
Closes reanahub/reana-client#497.

### DIFF
--- a/reana_workflow_controller/config.py
+++ b/reana_workflow_controller/config.py
@@ -11,6 +11,7 @@
 import os
 
 from reana_commons.config import REANA_COMPONENT_PREFIX, SHARED_VOLUME_PATH
+from reana_db.models import JobStatus
 
 from reana_workflow_controller.version import __version__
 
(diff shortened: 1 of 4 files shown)


reana_workflow_controller/config.py
```diff
@@ -25,7 +26,12 @@
 WORKFLOW_TIME_FORMAT = "%Y-%m-%dT%H:%M:%S"
 """Time format for workflow starting time, created time etc."""
 
-PROGRESS_STATUSES = ["running", "finished", "failed", "total"]
+PROGRESS_STATUSES = [
+    ("running", JobStatus.running),
+    ("finished", JobStatus.finished),
+    ("failed", JobStatus.failed),
+    ("total", None),
+]
```
ParthS007, line 34: Minor: Is there any reason why we opted for `tuple` list instead of `dict` of `PROGRESS_STATUSES`?

reana_workflow_controller/consumer.py
```diff
@@ -144,7 +144,7 @@ def _update_run_progress(workflow_uuid, msg):
     job_progress = workflow.job_progress
     if "cached" in msg["progress"]:
         cached_jobs = msg["progress"]["cached"]  # noqa: F841
-    for status in PROGRESS_STATUSES:
+    for status, _ in PROGRESS_STATUSES:
```
ParthS007, line 147: Minor: We can keep `job_status` in place of `_`.

ParthS007 (approved):
Thanks for the PR. Tested locally, working fine 🚀

mvidalgarcia (approved):
Haven't tested it. Code looks good 👍

reana_workflow_controller/config.py
diegodelemos: This variable has been a workaround from the beginning, the solution would be to go for https://github.com/reanahub/reana-workflow-controller/issues/204#issuecomment-829033572. Once the described change would be done, this mapping would disappear, so no need to make it complicated.

reana_workflow_controller/consumer.py
diegodelemos: Why? the second value of the tuple is not used, so for clarity, I remove any importance by assigning it to `_`. The rest remains unchanged.

PR #380, opened 2021-05-06
list: add --include-progress and --include-workspace-size

ParthS007:
performance optimisations for list command

closes https://github.com/reanahub/reana-client/issues/510

### DIFF
--- a/reana_workflow_controller/rest/workflows.py
+++ b/reana_workflow_controller/rest/workflows.py
@@ -109,10 +109,12 @@ def get_workflows(paginate=None):  # noqa
         - name: include_progress
           in: query
           description: Include progress information of the workflows.
+          required: false
           type: boolean
         - name: include_workspace_size
           in: query
           description: Include size information of the workspace.
+          required: false
           type: boolean
       responses:
         200:
--- a/reana_workflow_controller/rest/workflows.py
+++ b/reana_workflow_controller/rest/workflows.py
@@ -106,6 +106,14 @@ def get_workflows(paginate=None):  # noqa
           description: Number of results per page (pagination).
           required: false
           type: integer
+        - name: include_progress
+          in: query
+          description: Include progress information of the workflows.
+          type: boolean
+        - name: include_workspace_size
+          in: query
+          description: Include size information of the workspace.
+          type: boolean
       responses:
         200:
           description: >-
@@ -217,6 +225,8 @@ def get_workflows(paginate=None):  # noqa
         sort = request.args.get("sort", "desc")
         search = request.args.get("search", "")
         status_list = request.args.get("status", "")
+        include_progress = request.args.get("include_progress", verbose)
+        include_workspace_size = request.args.get("include_workspace_size", verbose)
         if not user:
             return jsonify({"message": "User {} does not exist".format(user_uuid)}), 404
         workflows = []
@@ -237,7 +247,9 @@ def get_workflows(paginate=None):  # noqa
                 "status": workflow.status.name,
                 "user": user_uuid,
                 "created": workflow.created.strftime(WORKFLOW_TIME_FORMAT),
-                "progress": get_workflow_progress(workflow),
+                "progress": get_workflow_progress(
+                    workflow, include_progress=include_progress
+                ),
             }
             if type_ == "interactive" or verbose:
                 int_session = workflow.sessions.first()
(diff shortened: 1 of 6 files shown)


reana_workflow_controller/rest/workflows.py
```diff
@@ -217,6 +225,8 @@ def get_workflows(paginate=None):  # noqa
         sort = request.args.get("sort", "desc")
         search = request.args.get("search", "")
         status_list = request.args.get("status", "")
+        include_progress = request.args.get("include_progress", "")
+        include_workspace_size = request.args.get("include_workspace_size", "")
```
mvidalgarcia, line 229: Here it'd be safer to have `False` as the default value instead of an empty string.

reana_workflow_controller/rest/workflows.py
```diff
@@ -237,7 +247,19 @@ def get_workflows(paginate=None):  # noqa
...
+                "run_started_at": (
+                    workflow.run_started_at.strftime(WORKFLOW_TIME_FORMAT)
+                    if workflow.run_started_at
+                    else None
+                ),
+                "run_finished_at": (
+                    workflow.run_finished_at.strftime(WORKFLOW_TIME_FORMAT)
+                    if workflow.run_finished_at
+                    else None
+                ),
```
mvidalgarcia, line 259: These two keys used to be under `progress` as far as I understood. So it means that the API response format changes, couldn't this have some unforeseen incompatibilities in the clients (r-client, r-ui, external clients (lukas), etc.)?

reana_workflow_controller/rest/workflows.py
```diff
@@ -237,7 +247,19 @@ def get_workflows(paginate=None):  # noqa
...
+                "progress": (
+                    get_workflow_progress(workflow) if include_progress else {}
```
mvidalgarcia, line 261: If `include_progress` is True, would it mean that `run_started_at` and `run_finished_at` information is repeated at two different levels?

mvidalgarcia: I guess `verbose` should be checked here as well to keep backward compatibility?

ParthS007: Fixed the repetition.

reana_workflow_controller/rest/workflows.py
ParthS007: I have updated the PR, now both the keys are in `progress` which will not have API response format mismatch.

reana_workflow_controller/rest/workflows.py
```diff
@@ -252,7 +262,7 @@ def get_workflows(paginate=None):  # noqa
                 "human_readable": "",
                 "raw": -1,
             }
-            if verbose:
+            if include_workspace_size or verbose:
```
mvidalgarcia, line 265: I think here `or verbose` is not needed due to how you get `include_workspace_size` value.

```
include_workspace_size = request.args.get("include_workspace_size", verbose)
```

- If `--include-workspace-size` is passed, `include_workspace_size` will be `True`
- If `--verbose` is passed, `include_workspace_size` will be `True` as well.
- If none is passed, `include_workspace_size` will be `False` as `verbose` will be `False`.

reana_workflow_controller/rest/workflows.py
```diff
@@ -237,7 +247,7 @@ def get_workflows(paginate=None):  # noqa
                 "status": workflow.status.name,
                 "user": user_uuid,
                 "created": workflow.created.strftime(WORKFLOW_TIME_FORMAT),
-                "progress": get_workflow_progress(workflow),
+                "progress": get_workflow_progress(workflow, include_progress),
```
mvidalgarcia, line 250:
```suggestion
                "progress": get_workflow_progress(workflow, include_progress=include_progress),
```

For the sake of readability.

reana_workflow_controller/rest/workflows.py
ParthS007: Thanks for the catch. Fixed 👍

setup.py
```diff
@@ -46,7 +46,7 @@
 
 install_requires = [
     "Flask>=0.12",
-    "werkzeug>=0.16.1,<0.17",
+    "werkzeug>=2.0",
```
mvidalgarcia, line 49: Any particular reason why you change this? 
By the way, this is not effective unless you update `requirements.txt` as well.

PR #383, opened 2021-06-07
fix: add raw and human_readable format in size object

ParthS007:
closes https://github.com/reanahub/reana-client/issues/520

### DIFF
--- a/reana_workflow_controller/rest/utils.py
+++ b/reana_workflow_controller/rest/utils.py
@@ -479,16 +479,24 @@ def list_files_recursive_wildcard(directory_path, path):
        error messages.
     """
     paths, posix_dir_prefix = get_files_recursive_wildcard(directory_path, path)
-    return [
-        {
-            "name": str(path.relative_to(posix_dir_prefix)),
-            "size": path.stat().st_size,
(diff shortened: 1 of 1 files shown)


tiborsimko (approved):
Works well :+1: 

One comment might be to calculate `path.stat().st_size` only once, before the return statement, so that we don't do it twice... This may be perhaps interesting in case of very large files? _(micro-optimisation)_

ParthS007:
Thanks for the catch. Fixed it 👍

reana_workflow_controller/rest/utils.py
```diff
@@ -479,16 +479,22 @@ def list_files_recursive_wildcard(directory_path, path):
...
-            "size": path.stat().st_size,
-            "last-modified": datetime.fromtimestamp(path.stat().st_mtime).strftime(
-                WORKFLOW_TIME_FORMAT
-            ),
-        }
-        for path in paths
-    ]
+    for path in paths:
+        raw_size = path.stat().st_size
+        return [
```
tiborsimko, line 484: Beware, we cannot `return` from within the `for` loop, because this would show only one file and not continue with all the matched ones, for example if a user asks for `reana-client ls '*/*'`.

PR #390, opened 2021-07-20
workflow-run-manager: support snakemake

mvidalgarcia:
closes reanahub/reana-client#531

### DIFF
--- a/reana_workflow_controller/workflow_run_manager.py
+++ b/reana_workflow_controller/workflow_run_manager.py
@@ -69,6 +69,7 @@ from reana_workflow_controller.config import (  # isort:skip
     REANA_KUBERNETES_JOBS_MEMORY_LIMIT,
     REANA_WORKFLOW_ENGINE_IMAGE_CWL,
     REANA_WORKFLOW_ENGINE_IMAGE_SERIAL,
+    REANA_WORKFLOW_ENGINE_IMAGE_SNAKEMAKE,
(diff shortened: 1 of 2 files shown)


reana_workflow_controller/workflow_run_manager.py
```diff
@@ -121,6 +122,18 @@ class WorkflowRunManager:
             ),
             "environment_variables": WORKFLOW_ENGINE_COMMON_ENV_VARS,
         },
+        "snakemake": {
+            "image": "{}".format(REANA_WORKFLOW_ENGINE_IMAGE_SNAKEMAKE),
+            "command": (
+                "run-snakemake-workflow "
+                "--workflow-uuid {id} "
+                "--workflow-workspace {workspace} "
+                "--workflow-file '{workflow_file}' "
```
audrium, line 131: There is no way to get json representation of a snakefile like for the other engines? Or it is not implemented yet? If the later, I guess we could still add `"--workflow-json '{workflow_json}' "` for it to work in the future

audrium (approved):

reana_workflow_controller/workflow_run_manager.py
mvidalgarcia: This part doesn't seem trivial as the Snakefile is not a YAML file as in other workflow engines, what is more, it's even possible to have python code embedded. So I don't think it's going to be an easy task... 
I guess this is going to be needed for the complexity estimation? We can later check how to solve it.

PR #391, opened 2021-07-22
models: modify workspace_path to store full path

marcdiazsan:
Quit the dependency of upload-file,remove-file, list-files,mv-files and download, relying directly on the full-path stored in the database

Closes reanahub/reana-db#94

### DIFF
--- a/reana_workflow_controller/rest/utils.py
+++ b/reana_workflow_controller/rest/utils.py
@@ -19,6 +19,7 @@ import subprocess
 import traceback
 import time
 import zipfile
+import shutil
 from collections import OrderedDict
 from datetime import datetime
 from functools import wraps
@@ -29,7 +30,7 @@ from flask import current_app as app
 from flask import jsonify, request, send_file, send_from_directory
 from git import Repo
 from kubernetes.client.rest import ApiException
-from reana_commons.config import REANA_WORKFLOW_UMASK
+from reana_commons.config import REANA_WORKFLOW_UMASK, SHARED_VOLUME_PATH
 from reana_commons.k8s.secrets import REANAUserSecretsStore
 from reana_commons.utils import (
     get_workflow_status_change_verb,
@@ -195,9 +196,7 @@ def remove_workflow_jobs_from_cache(workflow):
     """
     jobs = Session.query(Job).filter_by(workflow_uuid=workflow.id_).all()
     for job in jobs:
-        job_path = remove_upper_level_references(
-            os.path.join(workflow.workspace_path, "..", "archive", str(job.id_))
-        )
+        job_path = os.path.join(workflow.workspace_path, "..", "archive", str(job.id_))
         Session.query(JobCache).filter_by(job_id=job.id_).delete()
         remove_workflow_workspace(job_path)
     Session.commit()
@@ -322,8 +321,7 @@ def create_workflow_workspace(
     :return: Absolute workspace path.
     """
     os.umask(REANA_WORKFLOW_UMASK)
-    reana_fs = fs.open_fs(app.config["SHARED_VOLUME_PATH"])
-    reana_fs.makedirs(path, recreate=True)
+    os.makedirs(path, exist_ok=True)
     if git_url and git_ref:
         secret_store = REANAUserSecretsStore(user_id)
         gitlab_access_token = secret_store.get_secret_value("gitlab_access_token")
@@ -331,10 +329,7 @@ def create_workflow_workspace(
             gitlab_access_token, REANA_GITLAB_HOST, git_url
         )
         repo = Repo.clone_from(
-            url=url,
-            to_path=os.path.abspath(reana_fs.root_path + "/" + path),
-            branch=git_branch,
-            depth=1,
+            url=url, to_path=os.path.abspath(path), branch=git_branch, depth=1,
         )
         repo.head.reset(commit=git_ref)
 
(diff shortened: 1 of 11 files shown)


reana_workflow_controller/rest/utils.py
```diff
@@ -345,18 +346,19 @@ def remove_workflow_workspace(path):
...
-    absolute_workspace_path = os.path.join(
-        app.config["SHARED_VOLUME_PATH"], workflow.workspace_path
-    )
+    absolute_workspace_path = workflow.workspace_path
+    # os.path.join(
+    # current_app.config["SHARED_VOLUME_PATH"], workflow.workspace_path
+    # )
     absolute_source_path = os.path.join(
-        app.config["SHARED_VOLUME_PATH"], absolute_workspace_path, source
+        SHARED_VOLUME_PATH, absolute_workspace_path, source
```
audrium, line 361: I think we don't need shared path to be mounted here?
isn't `absolute_source_path = os.path.join(workflow.workspace_path, source)` enough?

reana_workflow_controller/rest/utils.py
```diff
@@ -398,6 +400,7 @@ def list_directory_files(directory, search=None):
                         ResourceUnit.bytes_, file_details.size
                     ),
                 ),
+                "path": directory,
```
audrium, line 403: Maybe I miss something, why this line is needed here?

reana_workflow_controller/rest/utils.py
```diff
@@ -600,35 +603,33 @@ def get_workspace_diff(workflow_a, workflow_b, brief=False, context_lines=5):
...
+    # reana_fs = fs.open_fs(app.config["SHARED_VOLUME_PATH"])
+    if os.path.exists(workspace_a) and os.path.exists(workspace_b):
         diff_command = [
             "diff",
             "--unified={}".format(context_lines),
             "-r",
-            reana_fs.getospath(workspace_a),
-            reana_fs.getospath(workspace_b),
+            bytes(workspace_a, "utf-8"),
+            bytes(workspace_b, "utf-8"),
```
audrium, line 613: I think `bytes()` is not really needed here, since we use directories as strings anyway

reana_workflow_controller/rest/workflows.py
```diff
@@ -409,6 +411,9 @@ def create_workflow():  # noqa
             logs="",
             git_ref=git_ref,
             git_repo=git_repo,
+            workspace_path=SHARED_VOLUME_PATH
+            + "/"
+            + build_workspace_path(request.args["user"], workflow_uuid),
```
audrium, line 416: We should better change `build_workspace_path()` in `reana-db` to append shared path while creating the workflow

audrium:
You could mark PR as a draft if it's still work in progress. Also, please change PR description and add "closes reanahub/reana-db#94" to the description and commits

reana_workflow_controller/rest/utils.py
marcdiazsan: Yes, the shared path is not needed in this case.

reana_workflow_controller/rest/utils.py
marcdiazsan: It isn't, I added it to print the path of the files to check they where stored in the correct place. But is a better idea to check it making the query to the database. I have already deleted it

reana_workflow_controller/rest/utils.py
marcdiazsan: Yes, correct

reana_workflow_controller/rest/utils.py
```diff
@@ -345,19 +346,18 @@ def remove_workflow_workspace(path):
     :param path: Relative path to workspace directory.
     :return: None.
     """
-    reana_fs = fs.open_fs(app.config["SHARED_VOLUME_PATH"])
-    if reana_fs.exists(path):
+    # reana_fs = fs.open_fs(app.config["SHARED_VOLUME_PATH"])
+    if os.path.isdir(path):
         reana_fs.removetree(path)
```
audrium, line 351: Probably `os.rmdir()` could be used here instead of `reana_fs.removetree()`

reana_workflow_controller/rest/utils.py
```diff
@@ -193,10 +193,11 @@ def remove_workflow_jobs_from_cache(workflow):
     :param workflow: The workflow object that spawned the jobs.
     :return: None.
     """
+    workspace_rel_path = os.path.relpath(workflow.workspace_path, SHARED_VOLUME_PATH)
```
audrium, line 196: Why this line is needed? Maybe I'm missing something, but I think it will try to `remove_workflow_workspace()` without `/var/reana` which should be incorrect for some workspaces

reana_workflow_controller/rest/workflows.py
```diff
@@ -409,6 +411,7 @@ def create_workflow():  # noqa
             logs="",
             git_ref=git_ref,
             git_repo=git_repo,
+            workspace_path=build_workspace_path(request.args["user"], workflow_uuid),
```
audrium, line 414: I guess this line is here for testing purposes? I think it should be better to leave this part as it was before, since `build_workspace_path()` will be called during workflow initialization from `reana-db models`

reana_workflow_controller/rest/workflows_workspace.py
```diff
@@ -255,9 +254,10 @@ def download_file(workflow_id_or_name, file_name):  # noqa
 
         workflow = _get_workflow_with_uuid_or_name(workflow_id_or_name, user_uuid)
 
-        absolute_workflow_workspace_path = os.path.join(
-            current_app.config["SHARED_VOLUME_PATH"], workflow.workspace_path
-        )
+        absolute_workflow_workspace_path = workflow.workspace_path
```
audrium, line 257: I think here you could simply use `workflow.workspace_path` while calling `download_files_recursive_wildcard()`

reana_workflow_controller/rest/workflows_workspace.py
```diff
@@ -344,9 +344,10 @@ def delete_file(workflow_id_or_name, file_name):  # noqa
             return jsonify({"message": "User {} does not exist".format(user)}), 404
 
         workflow = _get_workflow_with_uuid_or_name(workflow_id_or_name, user_uuid)
-        abs_path_to_workspace = os.path.join(
-            current_app.config["SHARED_VOLUME_PATH"], workflow.workspace_path
-        )
+        abs_path_to_workspace = workflow.workspace_path
```
audrium, line 347: Similarly we could get rid of `abs_path_to_workspace` here as well. Same for other places

audrium (commented):
Looks good! :rocket: 

There are still some places where shared path is joined to the worspace like [here](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/k8s.py#L272), [here](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/workflow_run_manager.py#L627) and [here](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/rest/utils.py#L325)

Also this PR needs a rebase on top of recently updated master

reana_workflow_controller/rest/utils.py
marcdiazsan: You are right, as I modified the path in remove_workflow_workspace()

reana_workflow_controller/rest/utils.py
marcdiazsan: As removetree(path) removes the directory and it's contents and os.rmdir() only removes empty directories, I think shutils.rmtree(path) is a better alternative

tests/test_views.py
```diff
@@ -93,16 +93,20 @@ def test_get_workflows_missing_user(app):
...
             url_for("workflows.create_workflow"),
-            query_string={"user": default_user.id_},
+            query_string={
+                "user": default_user.id_,
+                "workspace_root_path": tmp_shared_volume_path,
+            },
             content_type="application/json",
             data=json.dumps(cwl_workflow_with_name),
         )
+        print(res.get_data(as_text=True))
```
audrium, line 109: Very minor: I guess we can remove this print statement as well as the one in line 267

reana_workflow_controller/rest/workflows.py
```diff
@@ -409,6 +417,14 @@ def create_workflow():  # noqa
             logs="",
             git_ref=git_ref,
             git_repo=git_repo,
+<<<<<<< HEAD
```
tiborsimko, line 420: Note this is a remainer from merge conflicts that must be solved.

tiborsimko (approved):

PR #393, opened 2021-07-30
workflow_run_manager: separated workspace volume from shared volume

audrium:
Goes on top of https://github.com/reanahub/reana-workflow-controller/pull/391


tiborsimko:
Merged as part of #391.

PR #400, opened 2021-09-21
workflows: filter workflows listed using -w workflow passed

marcdiazsan:
Closes reanahub/reana-client#452

### DIFF
--- a/reana_workflow_controller/rest/workflows.py
+++ b/reana_workflow_controller/rest/workflows.py
@@ -118,6 +118,11 @@ def get_workflows(paginate=None):  # noqa
           description: Include size information of the workspace.
           required: false
           type: boolean
+        - name: workflow_id_or_name
+          in: query
+          description: Optional analysis UUID or name to filter.
+          required: false
+          type: string
       responses:
         200:
           description: >-
@@ -231,6 +236,7 @@ def get_workflows(paginate=None):  # noqa
         status_list = request.args.get("status", "")
         include_progress = request.args.get("include_progress", verbose)
         include_workspace_size = request.args.get("include_workspace_size", verbose)
+        workflow = request.args.get("workflow_id_or_name")
         if not user:
             return jsonify({"message": "User {} does not exist".format(user_uuid)}), 404
         workflows = []
@@ -242,6 +248,8 @@ def get_workflows(paginate=None):  # noqa
         if status_list:
             workflow_status = [RunStatus[status] for status in status_list.split(",")]
             query = query.filter(Workflow.status.in_(workflow_status))
+        if workflow:
+            query = query.filter(Workflow.name == workflow)
         if sort not in ["asc", "desc"]:
             sort = "desc"
         column_sorted = getattr(Workflow.created, sort)()
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -84,6 +84,13 @@
             "name": "include_workspace_size",
             "required": false,
             "type": "boolean"
+          },
+          {
+            "description": "Optional analysis UUID or name to filter.",
+            "in": "query",
+            "name": "workflow_id_or_name",
+            "required": false,
+            "type": "string"
           }
         ],
         "produces": [


reana_workflow_controller/rest/workflows.py
```diff
@@ -231,6 +236,7 @@ def get_workflows(paginate=None):  # noqa
         status_list = request.args.get("status", "")
         include_progress = request.args.get("include_progress", verbose)
         include_workspace_size = request.args.get("include_workspace_size", verbose)
+        workflow = request.args.get("workflow_id_or_name")
```
audrium, line 239: I'd expect that `reana-client list -w workflow.42` would also work in the same way as in other commands and return the the status of a specific workflow run:
```shell
reana-client list -w workflow.1
NAME       RUN_NUMBER   CREATED               STARTED               ENDED                 STATUS
workflow   1            2021-09-22T08:27:38   2021-09-22T08:27:49   2021-09-22T08:27:56   finished

reana-client list -w workflow.2
NAME       RUN_NUMBER   CREATED               STARTED               ENDED                 STATUS
workflow   2            2021-09-22T08:29:41   2021-09-22T08:29:52   2021-09-22T08:29:59   finished

reana-client list -w workflow # IMO should return the latest workflow run
NAME       RUN_NUMBER   CREATED               STARTED               ENDED                 STATUS
workflow   2            2021-09-22T08:29:41   2021-09-22T08:29:52   2021-09-22T08:29:59   finished
```
Right now workflow name provided together with run number does not work. In order to make it work we would need to use `_get_workflow_with_uuid_or_name()` function which splits the string and returns the correct workflow which could be later compared by `id_`. I mean something like this:
```python
workflow = request.args.get("workflow_id_or_name")
workflow = (
    _get_workflow_with_uuid_or_name(workflow, user_uuid) if workflow else None
)
....
if workflow:
    query = query.filter(Workflow.id_ == workflow.id_)
```

This is a bit contradictory with last example of the [issue](https://github.com/reanahub/reana-client/issues/452):
```
$ reana-client list -w myanalysis
NAME                  RUN_NUMBER   CREATED               STARTED               ENDED                 STATUS
myanalysis           44            2020-12-08T09:35:15   2020-12-08T09:35:16   -                     running
myanalysis           43            2020-12-08T09:34:15   2020-12-08T09:34:16   2020-12-08T09:33:47   finished
myanalysis           42            2020-12-08T09:33:15   2020-12-08T09:33:16   2020-12-08T09:33:47   failed
```

@tiborsimko , can you clarify which would be the correct approach?

mvidalgarcia: My two cents. IMO, it's not very consistent allowing `reana-client list -w workflow.42` to just display the run 42 of this workflow, we already have `reana-client status` for this very same purpose.

I pictured the functionality described in the issue as "I want to see all the runs of my workflow called `foo`".

OTOH, we have commands like `logs` where one can just specify the workflow name (without the run number) and get the logs of the **last workflow run**, so the above behavior would bring be some inconsistency...

Maybe the `-w` is not needed in `reana-client list`? One can achieve the same results with `--filter`.

tiborsimko: >  IMO should return the latest workflow run

Yes, that would seem nicely consistent with the rest of commands such as `ls` or `status`... Perhaps we can do the same for `list`. But I guess for "list" people are usually expecting seeing all, so it could also be confusing to show only the last one, especially since we don't have `--include-previous`. Hence I thought that `list` could be different... We can muse IRL about respective pros/cons, or whether we harmonise further.

> we already have reana-client status for this very same purpose.

Not fully: `status` is only about statuses, whilst `list` can also display workspace size.  So it's not fully the same. That said, we could perhaps include `--include-workspace-size` for `status` command; that might be a nicer solution.

>  One can achieve the same results with --filter.

Filtering works via substrings, so one could type `list --filter k` to see all workflows containing `k`. Whilst `list -w k` would be exact match, so only listing workflows of that name. A bit like `status -w k`.  That sort of difference...

marcdiazsan: I agree that listing just the last run is not as useful for the `list` command, it's more natural to include them all. Adding `--include-workspace-size` for `status` could be a nice addition to solve showing only the last run

mvidalgarcia: Do we have a final decision about this? What is the expected behavior? @marcdiazsan @tiborsimko

marcdiazsan: We still don´t have a final decision. I still think that having the command as it is showing all the runs of a workflow is more natural for `list` and although is a little different from the other commands ´-w´ is a better behavior

mvidalgarcia (approved):
Listing all the workflow runs works as expected 👍

```console
$ reana-client list -w foo                         
NAME   RUN_NUMBER   CREATED               STARTED               ENDED                 STATUS  
foo    2            2021-09-28T14:06:01   -                     -                     created 
foo    1            2021-09-28T13:48:28   2021-09-28T13:48:40   2021-09-28T13:48:46   finished
```

Tests passing locally with [`reana-commons` PR](https://github.com/reanahub/reana-commons/pull/296) installed ✅ 

PS. Needs rebase

reana_workflow_controller/rest/workflows.py
mvidalgarcia: OK, let's go for that and we can re-evaluate it later.

PR #407, opened 2021-10-20
Upgrade interactive sessions to support networking.k8s.io/v1 API

VMois:
closes https://github.com/reanahub/reana/issues/482

Tests are failing, most probably, because `reana-commons` is also updated.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 Changes
 =======
 
+Version 0.9.0 (UNRELEASED)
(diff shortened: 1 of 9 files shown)


tiborsimko (commented):
Works well for Kubernetes 1.23 :+1: 

Please update `CHANGES.rst` as in the other two repositories. We can then release `r-commons` and upgrade the version here for the CI tests to pass.

tiborsimko (approved):

PR #413, opened 2021-10-29
consumer: set prefetch_count to limit received messages 

audrium:
closes https://github.com/reanahub/reana-server/issues/410

As described in https://github.com/reanahub/reana-commons/pull/314#issuecomment-954770879, by default messages in the queue may not spend any time at all and may be consumed straight away. In this case the priority queue will not get any opportunity to prioritize them. 

To illustrate with roofit example. Without the PR all the messages are sent to the consumer (unacked)
Screenshot 2021-10-29 at 16-29-02 RabbitMQ Management

With this PR only one message at a time is being processed in the consumer while the others wait in the queue (ready):
Screenshot 2021-10-29 at 16-32-11 RabbitMQ Management

Please note that this setting can slow down the job status processing, so we should be careful with selecting the proper value. Perhaps the value should be increased after some initial tests are done.

### DIFF
--- a/reana_workflow_controller/config.py
+++ b/reana_workflow_controller/config.py
@@ -48,6 +48,12 @@ SHARED_FS_MAPPING = {
 }
 """Mapping from the shared file system backend to the job file system."""
 
+REANA_JOB_STATUS_CONSUMER_PREFETCH_COUNT = int(
+    os.getenv("REANA_JOB_STATUS_CONSUMER_PREFETCH_COUNT", 200)
+)
+"""The value defines the max number of unacknowledged deliveries that are
+permitted on a ``jobs-status`` consumer."""
+
 REANA_WORKFLOW_ENGINE_IMAGE_CWL = os.getenv(
     "REANA_WORKFLOW_ENGINE_IMAGE_CWL", "reanahub/reana-workflow-engine-cwl:latest"
 )
--- a/reana_workflow_controller/consumer.py
+++ b/reana_workflow_controller/consumer.py
@@ -39,6 +39,7 @@ from reana_workflow_controller.config import (
     PROGRESS_STATUSES,
     REANA_GITLAB_URL,
     REANA_HOSTNAME,
+    REANA_JOB_STATUS_CONSUMER_PREFETCH_COUNT,
 )
 from reana_workflow_controller.errors import REANAWorkflowControllerError
 
@@ -64,6 +65,9 @@ class JobStatusConsumer(BaseConsumer):
                 queues=self.queue,
                 callbacks=[self.on_message],
                 accept=[self.message_default_format],
+                prefetch_count=REANA_JOB_STATUS_CONSUMER_PREFETCH_COUNT
+                if REANA_JOB_STATUS_CONSUMER_PREFETCH_COUNT
+                else None,
             )
         ]
 


audrium:
> Details on [prefetch affects on performance](https://www.rabbitmq.com/confirms.html#channel-qos-prefetch-throughput) from RabbitMQ docs.

Perhaps even if we abandon [this PR](https://github.com/reanahub/reana-commons/pull/314) because of lack of effect, we should still use the `prefetch_count` in `jobs-status` queue with some safe value (100-300 range). According to the [docs](https://www.rabbitmq.com/confirms.html#channel-qos-prefetch-throughput):
```
Consumers that consume a lot of messages without acknowledging will lead to memory consumption growth on the node they are connected to. Finding a suitable prefetch value is a matter of trial and error and will vary from workload to workload. Values in the 100 through 300 range usually offer optimal throughput and do not run significant risk of overwhelming consumers.
```
On heavy load without using the  `prefetch_count` all the messages would go straight to consumer and could increase RAM consumption.

audrium:
Made `prefetch_count` value parameterizable via environment variable with a default value of 200

reana_workflow_controller/config.py
```diff
@@ -48,6 +48,12 @@
 }
 """Mapping from the shared file system backend to the job file system."""
 
+REANA_JOB_STATUS_CONSUMER_PREFETCH_COUNT = int(
+    os.getenv("REANA_JOB_STATUS_CONSUMER_PREFETCH_COUNT", 200)
+)
+"""Prefetch count value defines max number of messages that will be delivered to
+the ``jobs-status`` queue consumer."""
```
VMois, line 55: *nitpick*: will be good to make it clear that messages that are prefetched are unacknowledged 

```suggestion
"""The value defines the max number of unacknowledged deliveries that are permitted on a ``jobs-status`` consumer."""
```

Taken from [RabbitMQ docs](https://www.rabbitmq.com/confirms.html#channel-qos-prefetch)

VMois (commented):
LGTM. Just a nitpick comment.

reana_workflow_controller/consumer.py
```diff
@@ -64,6 +65,7 @@ def get_consumers(self, Consumer, channel):
                 queues=self.queue,
                 callbacks=[self.on_message],
                 accept=[self.message_default_format],
+                prefetch_count=REANA_JOB_STATUS_CONSUMER_PREFETCH_COUNT,
```
tiborsimko, line 68: Could you perhaps do `prefetch_count=ENV if ENV else False`  so that we could use ENV=0 as a way to switch off the prefetch count?

(I tested 1 and 200 and there did not seem to have been much visible difference...)

VMois: There will be no visible difference in terms of performance (value `200` vs unlimited) with a few workflows, even with a couple of hundreds, I think. Prefetch doesn't improve the performance of our code, jobs-status consumer will work as it was working before. It limits the buffer size so RAM consumption will not go crazy if we will launch 100k workflows and jobs-status consumer will suddenly prefetch 10k+ messages and eat RAM from the node.

tiborsimko (approved):
Confirmed that the PR has no visible effect on performance when running 1k (250 in parallel) workflows with prefetch 0, 1 or 200.

PR #417, opened 2021-11-03
consumer: fetch specific workflow pod during deletion

audrium:
closes https://github.com/reanahub/reana-job-controller/issues/326

Once the workflow finishes consumer tries to `_delete_workflow_engine_pod()`. Before it was fetching all available pods in the cluster via `current_k8s_corev1_api_client.list_namespaced_pod()` and iterating to find a specific pod to terminate. Now it will fetch the right pod using `label_selector` making it faster to terminate.

### DIFF
--- a/reana_workflow_controller/workflow_run_manager.py
+++ b/reana_workflow_controller/workflow_run_manager.py
@@ -438,7 +438,10 @@ class KubernetesWorkflowRunManager(WorkflowRunManager):
 
         workflow_metadata = client.V1ObjectMeta(
             name=name,
-            labels={"reana_workflow_mode": "batch"},
+            labels={
+                "reana_workflow_mode": "batch",
+                "reana-run-batch-workflow-uuid": str(self.workflow.id_),
+            },
             namespace=REANA_RUNTIME_KUBERNETES_NAMESPACE,
         )
         job = client.V1Job()
--- a/reana_workflow_controller/consumer.py
+++ b/reana_workflow_controller/consumer.py
@@ -267,6 +267,7 @@ def _delete_workflow_engine_pod(workflow):
     try:
         jobs = current_k8s_corev1_api_client.list_namespaced_pod(
             namespace=REANA_RUNTIME_KUBERNETES_NAMESPACE,
+            label_selector=f"reana-run-batch-workflow-uuid={str(workflow.id_)}",
         )
         for job in jobs.items:
             if str(workflow.id_) in job.metadata.name:


reana_workflow_controller/workflow_run_manager.py
```diff
@@ -438,7 +438,10 @@ def _create_job_spec(
 
         workflow_metadata = client.V1ObjectMeta(
             name=name,
-            labels={"reana_workflow_mode": "batch"},
+            labels={
+                "reana_workflow_mode": "batch",
+                "reana-run-batch-workflow-uuid": str(self.workflow.id_),
```
mvidalgarcia, line 443: Should we align on the naming [as here](https://github.com/reanahub/reana-job-controller/commit/9600b86e633d879a88e2a6c653dee248467562d4#diff-91c1d8d981e6114fea16cab4d9162666f97d596d1bcfc293f5a0b7e840b127b2R150) (probably your name is more complete) or do you think it might interfere? Maybe it's a good idea to keep `run-batch` and `run-jobs` pods with different label names?

tiborsimko (commented):
Works great :+1: 

Status before:

pMSSM-NoSys-1200Mi-10nodeb-56nodej-250max-21k8s-mqprio-1k_execution_progress

pMSSM-NoSys-1200Mi-10nodeb-56nodej-250max-21k8s-mqprio-1k_histogram_runtime


Status after:

pMSSM-NoSys-1200Mi-10nodeb-56nodej-250max-21k8s-mqprio-rwc417-1k_execution_progress

pMSSM-NoSys-1200Mi-10nodeb-56nodej-250max-21k8s-mqprio-rwc417-1k_histogram_runtime

Gains about 30 percent in the overall execution time, increases the steady-state burn down rate, and throughout the test I saw only a couple or so occurrences of NotReady pod states... So this solves nicely the core of the "blue spread" issue :tada:

reana_workflow_controller/workflow_run_manager.py
audrium: As we discussed in the standup, it's better to keep label keys different for `run-batch` and `run-jobs` pods for more precise pod selection. Perhaps we could rename `run-jobs` label keys accordingly: `reana-run-job-workflow-uuid`? WDYT? I could create a quick PR to change it

mvidalgarcia: `reana-run-job-workflow-uuid` looks good to me.

audrium: Addressed it [here](https://github.com/reanahub/reana-job-controller/pull/340)

mvidalgarcia (approved):
Tested together with https://github.com/reanahub/reana-job-controller/pull/340 by running `reana-dev run-example` and working fine.

PR #418, opened 2021-11-04
utils: optimize querying workflow progress information

### LINKED ISSUE #415: rest: get_workflows, job progress is queried for each workflow despite being not included in the response
In the `get_workflows` API when `include_progress` is not passed (= False), job progress is still queried from DB. 

Listing from [get_workflow_progress](https://github.com/reanahub/reana-workflow-controller/blob/d650d0c059ff17bd3b62d5e34addfa1af9e24c1c/reana_workflow_controller/rest/utils.py#L628) function:

Jobs are queried at the beginning:

https://github.com/reanahub/reana-workflow-controller/blob/d650d0c059ff17bd3b62d5e34addfa1af9e24c1c/reana_workflow_controller/rest/utils.py#L636-L641

But only used and *may* be included only here:

https://github.com/reanahub/reana-workflow-controller/blob/d650d0c059ff17bd3b62d5e34addfa1af9e24c1c/reana_workflow_controller/rest/utils.py#L658-L671

In case the user doesn't want to include progress, unnecessary queries are still performed. For 1000 workflows, it potentially increases response time a lot.

### LINKED ISSUE #416: get_current_job_progress is making a lot of unnecessary DB queries
It looks like [get_current_job_progress](https://github.com/reanahub/reana-workflow-controller/blob/d650d0c059ff17bd3b62d5e34addfa1af9e24c1c/reana_workflow_controller/rest/utils.py#L172) function is doing too many queries for each workflow:

https://github.com/reanahub/reana-workflow-controller/blob/d650d0c059ff17bd3b62d5e34addfa1af9e24c1c/reana_workflow_controller/rest/utils.py#L172-L188

After a single query to get *all* jobs, for each job from previous results a new query is performed to get the same job Assuming `id_` for each job is unique, even for retried ones, this is not optimal behavior.

Maybe, an additional job query has some purpose but, unfortunately, context is lost.

Originated from #415

mvidalgarcia:
- Only query progress information when `include_progress` is set
- Optimize querying the most recent job info

closes #415
closes #416

### DIFF
--- a/reana_workflow_controller/rest/utils.py
+++ b/reana_workflow_controller/rest/utils.py
@@ -25,8 +25,9 @@ from datetime import datetime
 from functools import wraps
 from io import BytesIO
 from pathlib import Path
+from typing import Dict
+from uuid import UUID
 
-from flask import current_app as app
 from flask import jsonify, request, send_file, send_from_directory
 from git import Repo
 from kubernetes.client.rest import ApiException
@@ -167,25 +168,6 @@ def build_workflow_logs(workflow, steps=None, paginate=None):
         all_logs[str(job.id_)] = item
 
     return all_logs
-
-
-def get_current_job_progress(workflow_id):
-    """Return job."""
-    current_job_commands = {}
-    workflow_jobs = Session.query(Job).filter_by(workflow_uuid=workflow_id).all()
-    for workflow_job in workflow_jobs:
-        job = (
-            Session.query(Job)
-            .filter_by(id_=workflow_job.id_)
-            .order_by(Job.created.desc())
-            .first()
-        )
-        if job:
-            current_job_commands[str(job.id_)] = {
-                "prettified_cmd": job.prettified_cmd,
-                "current_job_name": job.job_name,
-            }
-    return current_job_commands
 
 
 def remove_workflow_jobs_from_cache(workflow):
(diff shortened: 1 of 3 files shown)


reana_workflow_controller/rest/utils.py
```diff
@@ -625,20 +606,33 @@ def get_workspace_diff(workflow_a, workflow_b, brief=False, context_lines=5):
             )
 
 
-def get_workflow_progress(workflow, include_progress=False):
+def get_most_recent_job_info(workflow_id: UUID) -> dict:
```
VMois, line 609: `from typing import Dict`

```suggestion
def get_most_recent_job_info(workflow_id: UUID) -> Dict[str, str]:
```

[explanation why](https://stackoverflow.com/questions/39458193/using-list-tuple-etc-from-typing-vs-directly-referring-type-as-list-tuple-etc)

reana_workflow_controller/rest/utils.py
```diff
@@ -625,20 +606,33 @@ def get_workspace_diff(workflow_a, workflow_b, brief=False, context_lines=5):
...
+    )
+    if most_recent_job:
+        current_job_commands = {
+            "prettified_cmd": most_recent_job.prettified_cmd,
+            "current_job_name": most_recent_job.job_name,
+        }
+    return current_job_commands
+
+
+def get_workflow_progress(workflow: Workflow, include_progress: bool = False) -> dict:
```
VMois, line 626:
```suggestion
def get_workflow_progress(workflow: Workflow, include_progress: bool = False) -> Dict:
```

VMois (commented):
Works nice. Tested `list` + `--include-progress -v`, also `status` + `-v`. Left some small comments.

Did a small comparison of a speed locally. **100** workflows in DB (42 finished so have jobs, 59 just created so no jobs). I used `time reana-client list` command and did three runs. Results of `Real` time:

Before this PR: `3.6s`, `2.8s`, `3s`;
After this PR: `2.1s`, `1.4s`, `1.5s`

This is only 100 workflows, I assume improvements will be more noticeable with 1000+ workflows.

reana_workflow_controller/rest/utils.py
```diff
@@ -625,20 +607,33 @@ def get_workspace_diff(workflow_a, workflow_b, brief=False, context_lines=5):
             )
 
 
-def get_workflow_progress(workflow, include_progress=False):
+def get_most_recent_job_info(workflow_id: UUID) -> Dict[str, str]:
+    """Return most recent Job cmd and name from a certain workflow."""
+    current_job_commands = {}
+    most_recent_job = (
+        Session.query(Job)
```
audrium, line 614: nice improvement :+1:  one query instead of.. a lot :)

mvidalgarcia: The worse part was that this function was called always, independently of `verbose` or `include_progress` values 😬

VMois (approved):

PR #429, opened 2022-01-18
job status consumer: fix workflow engine logs when job is not deleted

### LINKED ISSUE #426: job-status consumer: engine logs are not stored when workflow fails and is not cleaned up
When a workflow fails but cleanup of failed workflows is disabled, logs from workflow engine container are not stored in DB, therefore `reana-client logs` will show an empty result. 

## How to reproduce?

In the dev environment, if workflow fails you will see no logs from the engine container. You might see one or two lines of error logs that could be generated by [this](https://github.com/reanahub/reana-commons/blob/c28e8c2d520bca2bd737c31dc46f2768fcbc00aa/reana_commons/workflow_engine.py#L191). But I assume we want engine container logs for all workflows.

## What is happening?

Reson is due to misleading `_delete_workflow_engine_pod` function. It contains code that records logs and deletes pod together despite its name containing only "delete". Because this function is only run when `RunStatus.should_cleanup_job` is true, we do not store the logs for `failed` workflows, or any workflow status if it is listed in `REANA_RUNTIME_KUBERNETES_KEEP_ALIVE_JOBS_WITH_STATUSES`.

https://github.com/reanahub/reana-workflow-controller/blob/a3e2af49a4c8153a46cee939ff74cfb81ae2c356/reana_workflow_controller/consumer.py#L149-L150

This situation is usually happening in the dev environment, but can also happen if the administrator configured `REANA_RUNTIME_KUBERNETES_KEEP_ALIVE_JOBS_WITH_STATUSES` value in the helm for some reason.

## Possible solution

`_delete_workflow_engine_pod` should do one thing only - delete engine pod. We should move code related to logs to some function like `_store_engine_logs`. This function should be called when workflow transitions to a final status (`finished`, `failed`) despite cleanup conditions.

VMois:
- separate deletion of workflow job and saving workflow engine pod logs, so workflow engine logs are captured even if workflow job is not deleted.

closes #426
closes reanahub/reana-workflow-engine-snakemake#35

## How to test

1. Try to reproduce the issue, it should not be possible.

### DIFF
--- a/reana_workflow_controller/consumer.py
+++ b/reana_workflow_controller/consumer.py
@@ -28,6 +28,7 @@ from reana_commons.utils import (
     calculate_file_access_time,
     calculate_hash_of_dir,
     calculate_job_input_hash,
+    build_unique_component_name,
 )
 from reana_db.database import Session
 from reana_db.models import Job, JobCache, Workflow, RunStatus
@@ -144,15 +145,27 @@ def _update_workflow_status(workflow, status, logs):
             _update_commit_status(workflow, status)
 
         if status not in ALIVE_STATUSES:
+            workflow.run_finished_at = datetime.now()
+            workflow.logs = workflow.logs or ""
+
             try:
-                workflow.run_finished_at = datetime.now()
-                if RunStatus.should_cleanup_job(status):
-                    _delete_workflow_engine_pod(workflow)
-            except REANAWorkflowControllerError:
+                workflow_engine_logs = _get_workflow_engine_pod_logs(workflow)
+                workflow.logs += workflow_engine_logs + "\n"
+            except REANAWorkflowControllerError as exception:
                 logging.error(
-                    f"Could not clean up workflow engine for workflow {workflow.id_}"
+                    f"Could not fetch workflow engine pod logs for workflow {workflow.id_}."
+                    f" Error: {exception}"
                 )
                 workflow.logs += "Workflow engine logs could not be retrieved.\n"
+
+            if RunStatus.should_cleanup_job(status):
+                try:
+                    _delete_workflow_job(workflow)
+                except REANAWorkflowControllerError as exception:
+                    logging.error(
+                        f"Could not clean up workflow job for workflow {workflow.id_}."
+                        f" Error: {exception}"
+                    )
 
 
 def _update_commit_status(workflow, status):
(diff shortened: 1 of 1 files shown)


reana_workflow_controller/consumer.py
```diff
@@ -144,16 +145,28 @@ def _update_workflow_status(workflow, status, logs):
...
-                    f"Could not clean up workflow engine for workflow {workflow.id_}"
+                    f"Could not fetch workflow engine pod logs for workflow {workflow.id_}."
+                    f" Error: {exception}"
                 )
                 workflow.logs += "Workflow engine logs could not be retrieved.\n"
 
+            if RunStatus.should_cleanup_job(status):
+                try:
+                    _delete_workflow_job(workflow)
+                except REANAWorkflowControllerError as exception:
```
VMois, line 164: When the `_delete_workflow_job` function fails and raises `REANAWorkflowControllerError`exception, we log the `Could not clean up workflow job for workflow...` message. And, when `REANAWorkflowControllerError` is raised in `_delete_workflow_job`, we have a message `Workflow engine pod could not be deleted...`. Those messages are basically the same. I feel like it is a bit overkill. What about just logging the exception itself? LIke:

```python
try:
  _delete_workflow_job(workflow)
except REANAWorkflowControllerError as exception:
  logging.error(exception)
```

audrium: Since you created dedicated functions `_get_workflow_engine_pod_logs()` and `_delete_workflow_job()` with `try except` blocks inside I think it's not worth to have it in `_update_workflow_status()` as well. All the logging could be done in dedicated functions and the main `_update_workflow_status()` would be simplified a bit, WDYT?

audrium (approved):
Nicely done! Works as expected

reana_workflow_controller/consumer.py
VMois: In the `except` block for workflow engine logs in `_update_workflow_status`, I have this part:

```python
try:
  ...
except REANAWorkflowControllerError as exception:
  ...
  workflow.logs += "Workflow engine logs could not be retrieved.\n"
```
If I will hide exception handling inside `_get_workflow_engine_pod_logs,  I will not be able to add to `workflow.logs`. I would prefer to still raise an exception inside the `_delete_workflow_job()` and `_get_workflow_engine_pod_logs()`. And handle them in `_update_workflow_status()`.

VMois: Ohh... It is merged already. I didn't add this change 😢

audrium: From your last comment I understood there is nothing to change. You were referring to the first comment perhaps? My bad, I guess, sorry

VMois: It is okay. It was not clear from my side that I want to add more stuff.

Would be good to adjust how we re-request review so things like this are not happening.

PR #434, opened 2022-03-10
rest: improve filtering by workflow in get workflows endpoint

audrium:
closes https://github.com/reanahub/reana-ui/issues/229

### DIFF
--- a/reana_workflow_controller/rest/workflows.py
+++ b/reana_workflow_controller/rest/workflows.py
@@ -12,7 +12,6 @@ import json
 from uuid import uuid4
 
 from flask import Blueprint, jsonify, request
-from reana_commons.config import SHARED_VOLUME_PATH
 from reana_db.database import Session
 from reana_db.utils import build_workspace_path
 from reana_db.models import (
@@ -36,6 +35,7 @@ from reana_workflow_controller.rest.utils import (
     get_workflow_name,
     get_workflow_progress,
     get_workspace_diff,
+    is_uuid_v4,
     use_paginate_args,
 )
 
@@ -241,7 +241,7 @@ def get_workflows(paginate=None):  # noqa
         status_list = request.args.get("status", "")
         include_progress = request.args.get("include_progress", verbose)
         include_workspace_size = request.args.get("include_workspace_size", verbose)
-        workflow = request.args.get("workflow_id_or_name")
+        workflow_id_or_name = request.args.get("workflow_id_or_name")
         if not user:
             return jsonify({"message": "User {} does not exist".format(user_uuid)}), 404
         workflows = []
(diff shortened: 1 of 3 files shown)


reana_workflow_controller/rest/workflows.py
```diff
@@ -254,7 +255,9 @@ def get_workflows(paginate=None):  # noqa
             workflow_status = [RunStatus[status] for status in status_list.split(",")]
             query = query.filter(Workflow.status.in_(workflow_status))
         if workflow:
```
mvidalgarcia, line 257: The `workflow` variable name here might be a bit confusing as it looks like an instance of the Workflow model, WDYT about keeping `workflow_id_or_name`?

IMHO, we should use marshmallow schemas [as we do for r-server](https://github.com/reanahub/reana-server/blob/b6cd672a94975124221da4875c17ebe50dd82ae5/reana_server/rest/workflows.py#L61) to avoid all the manual variable assignments of request args. It doesn't belong to this task though.

reana_workflow_controller/rest/workflows.py
```diff
@@ -254,7 +255,9 @@ def get_workflows(paginate=None):  # noqa
...
-            query = query.filter(Workflow.name == workflow)
+            query = query.filter(
+                or_(Workflow.name == workflow, Workflow.id_ == workflow)
```
mvidalgarcia, line 259: I think this breaks the filter by workflow name. For instance, this call (`https://localhost:30443/api/workflows?workflow_id_or_name=roofit`) is giving an error to me, can you reproduce the same? 

It looks like `query.filter(Workflow.id_ == workflow)` doesn't like comparisons with regular strings.

```python
>>> user.workflows.filter(Workflow.id_ == 'foo').count()
[StatementError: StatementError('(builtins.ValueError) bytes is not a 16-char string')](http://localhost:31984/debug/session/140507376465536)
```

mvidalgarcia:
PS. There's a tiny typo in the commit message, s/enpoint/endpoint/

reana_workflow_controller/rest/workflows.py
audrium: Indeed `workflow` variable name was confusing here, changed it to `workflow_id_or_name`.
Regarding the `marshmallow schemas`, totally agree! It would be nice to unify and use it in all the endpoints, perhaps worth to create an issue for that?

reana_workflow_controller/rest/workflows.py
audrium: Amazing find, thanks! Unexpected behavior while filtering uuids.. I've added a quick check to distinguish queries for uuids and workflow names

reana_workflow_controller/rest/workflows.py
mvidalgarcia: > Regarding the `marshmallow schemas`, totally agree! It would be nice to unify and use it in all the endpoints, perhaps worth to create an issue for that?

I wouldn't mind at all! However, we probably should [upgrade to Marshmallow 3](https://github.com/reanahub/reana-job-controller/issues/352) first.

reana_workflow_controller/rest/workflows.py
mvidalgarcia: Works well now! 👍 Strange behavior indeed.

mvidalgarcia (approved):
Works as expected! 👍

PR #443, opened 2022-04-04
job-status-consumer: improve logging of "not alive" workflows

### LINKED ISSUE #437: job-status-consumer: make it clear what status "not alive workflow" event has in logs, check and clean "not alive workflow" pods
Example of logs from `job-status-consumer`:

```
2022-03-24 14:20:16,312 | root | MainThread | WARNING | Event for not alive workflow 9b67170b-33ce-4dc3-8150-99e490afcade received:
{"workflow_uuid": "9b67170b-33ce-4dc3-8150-99e490afcade", "logs": "", "status": 1, "message": {"progress": {"total": {"total": 1, "job_ids": []}}}}
2022-03-24 14:20:16,315 | root | MainThread | WARNING | Event for not alive workflow 9b67170b-33ce-4dc3-8150-99e490afcade received:
{"workflow_uuid": "9b67170b-33ce-4dc3-8150-99e490afcade", "logs": "", "status": 1, "message": {"progress": {"running": {"total": 1, "job_ids": ["51060737-7e03-4255-a779-0e71e061e5f5"]}}}}
2022-03-24 14:20:16,318 | root | MainThread | WARNING | Event for not alive workflow 9b67170b-33ce-4dc3-8150-99e490afcade received:
```

Out of the logs above, there are two points:

- when an event is "no alive workflow" it is not clear what is the actual status of the workflow, "deleted" or maybe workflow doesn't exist at all; would be good to log the status of a workflow;

- it looks like if the workflow is deleted in the middle of the running (or when in a pending state) there is a chance that the workflow pod is still left out in Kubernetes; we need to check if any pod exists even in case of not alive workflow.

VMois:
addresses #437 

How to test:

1. Start helloworld workflow - `reana-client run -w test`
2. As soon as workflow starts, wait 3-5 seconds (it should go to the `pending` state) and delete it `reana-client delete -w test`
3. Check `reana-client list`, it will show you that `test` workflow is deleted
4. Check `kubectl get pods`, you will find batch pod in `NotReady` state
5. Check `kubectl logs deployment/reana-workflow-controller job-status-consumer`, you will see that workflow was not in an alive state in DB and logs indicate that. 

```
2022-04-04 11:27:58,215 | root | MainThread | WARNING | Event for not alive workflow 7c081fbc-42b2-4889-841d-a48ccd68dda2 with DB status RunStatus.deleted received:
{"workflow_uuid": "7c081fbc-42b2-4889-841d-a48ccd68dda2", "logs": "", "status": 2, "message": {"progress": {"finished": {"total": 1, "job_ids": ["17aeb786-69ec-4ddd-911a-542c59d7bf14"]}}}}
Ignoring...
```

This doesn't solve a core issue but, at least, we can have better logs.

### DIFF
--- a/reana_workflow_controller/consumer.py
+++ b/reana_workflow_controller/consumer.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2018, 2019, 2020, 2021 CERN.
+# Copyright (C) 2018, 2019, 2020, 2021, 2022 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -87,22 +87,19 @@ class JobStatusConsumer(BaseConsumer):
                 Session.query(Workflow)
                 .filter(
                     Workflow.id_ == workflow_uuid,
-                    Workflow.status.in_(ALIVE_STATUSES),
                 )
                 .one_or_none()
             )
-            if workflow:
+            if workflow and workflow.status in ALIVE_STATUSES:
                 next_status = body_dict.get("status")
                 if next_status:
                     next_status = RunStatus(next_status)
                     logging.info(
-                        " [x] Received workflow_uuid: {0} status: {1}".format(
-                            workflow_uuid, next_status
-                        )
+                        f" [x] Received workflow_uuid: {workflow_uuid} status: {next_status}"
                     )
 
-                logs = body_dict.get("logs") or ""
                 if workflow.can_transition_to(next_status):
+                    logs = body_dict.get("logs") or ""
                     _update_workflow_status(workflow, next_status, logs)
                     if "message" in body_dict and body_dict.get("message"):
                         msg = body_dict["message"]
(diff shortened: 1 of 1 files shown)


reana_workflow_controller/consumer.py
```diff
@@ -240,46 +232,6 @@ def _update_job_progress(workflow_uuid, msg):
                     job.status = job_status
 
 
-def _update_job_cache(msg):
```
tiborsimko, line 243: Hmm, whilst it is true that caching has been disabled since a long time, we may want to resurrect and retest it one day perhaps... Hence let me muse a bit on this. 

Does this code actually runs on the server-side? We have cache=off by default on the client side, so I think this code should not even be triggered on the server-side. If it is not, then we don't seem to have a problem here? If it is, then we should fix it so that it would not be? (or inactivate it via a server-side configuration settings without removal of the code).

That said, philosophically speaking, I like [suckless approach](https://suckless.org/philosophy/) arguing that _"the more code lines you have removed, the more progress you have made"_ :wink: Removing old inactive code is a good thing, especially if we investigate something like Armada or Kueue soon... Then it might be a perfect time for cleaning job cache everywhere. Until then, perhaps "inactivation" may still be having some advantages.

VMois: Honestly, there is a lack of documentation for cache internals so I have no idea how it works or where it is configured exactly. Need to dig into the code. 

Doesn't look like we have a cache config in the `helm` chart for this. Also, user docs only have a few things mentioned like the "CACHE=off" parameter for r-client. There are some Python tests for cache.

Looks like only the `serial` engine supports cache. Other engines do not check if a job is cached from the `job-controller`.

I heard that cache was not working so, in my opinion, we should remove it. Otherwise, this is just a dead code. But, I agree, maybe, I should create a separate issue to remove the cache and we can address it later (espcecially if we plan work on improving consumers and internals)

PR #446, opened 2022-04-14
rest: update disk usage after workspace deletion

mdonadoni:
Closes reanahub/reana-ui#259
Depends on reanahub/reana-db#168

How to test:
Trying different combinations of `disk` and `cpu` in `WORKFLOW_TERMINATION_QUOTA_UPDATE_POLICY`
1. Launch some workflows
2. Simulate a periodic quota update
```
kubectl create job --from cronjob/reana-resource-quota-update manual-quota-update-001
```
3. Delete one of the workflows, together with its workspace

Expected result: The disk quota is updated immediately after the deletion of a workspace, even if `disk` is not in `WORKFLOW_TERMINATION_QUOTA_UPDATE_POLICY`


audrium (approved):
LGMT, Great work!

PR #448, opened 2022-05-06
workflow_run_manager: setup Kerberos for workflow engine

mdonadoni:
Closes reanahub/reana-workflow-engine-snakemake#44


audrium (approved):
Nicely done!
Similarly to `r-j-c` we need to release a new version of `r-commons` for the CI to be green

PR #450, opened 2022-05-24
rest: sort workflow list endpoint by most quota used workflows

audrium:
closes https://github.com/reanahub/reana-ui/issues/270

To test:
- Change `termination_update_policy` to `disk,cpu` in [values.yaml](https://github.com/reanahub/reana/blob/master/helm/reana/values.yaml) and redeploy
- Run some workflows (manipulate `quota_used` values in `workflow_resource` db table if needed)
- Verify that sorting works correctly

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,7 @@ Version 0.9.0 (UNRELEASED)
 
 - Adds support for Kubernetes networking/v1 API to interactive sessions.
 - Changes workflow list endpoint to add the possibility to filter by workflow id.
+- Changes workflow list endpoint to add the possibility to sort workflows by most used disk and cpu quota.
 - Changes default consumer prefetch count to handle 10 messages instead of 200 in order to reduce the probability of 406 PRECONDITION errors on message acknowledgement.
 - Changes the workflow status endpoint to update the disk quota usage when a workspace is deleted.
 - Adds setup of Kerberos when executing a workflow engine.
--- a/reana_workflow_controller/rest/workflows.py
+++ b/reana_workflow_controller/rest/workflows.py
@@ -14,12 +14,9 @@ from uuid import uuid4
 from flask import Blueprint, jsonify, request
 from reana_db.database import Session
 from reana_db.utils import build_workspace_path
-from reana_db.models import (
-    User,
-    Workflow,
-    RunStatus,
-)
-from reana_db.utils import _get_workflow_with_uuid_or_name
+from reana_db.models import User, Workflow, RunStatus, ResourceType, WorkflowResource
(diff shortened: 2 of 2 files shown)


reana_workflow_controller/rest/workflows.py
```diff
@@ -259,9 +256,20 @@ def get_workflows(paginate=None):  # noqa
                 if is_uuid_v4(workflow_id_or_name)
                 else query.filter(Workflow.name == workflow_id_or_name)
             )
-        if sort not in ["asc", "desc"]:
-            sort = "desc"
-        column_sorted = getattr(Workflow.created, sort)()
+        column_sorted = Workflow.created.desc()
+        if sort in ["disk-desc", "cpu-desc"]:
+            resource_type = ResourceType[sort.split("-")[0]].name
```
mdonadoni, line 261: Isn't this the same as `resource_type = sort.split("-")[0]`?

reana_workflow_controller/rest/workflows.py
```diff
@@ -259,9 +256,20 @@ def get_workflows(paginate=None):  # noqa
...
+            resource = get_default_quota_resource(resource_type)
+            query = query.join(
+                WorkflowResource,
+                and_(
+                    Workflow.id_ == WorkflowResource.workflow_id,
+                    WorkflowResource.resource_id == resource.id_,
+                ),
+            )
```
mdonadoni, line 269: When ordering by most cpu/disk usage, workflows that do not have entries in the `workflow_resource` table are not returned by this endpoint. This can happen when the quota updates are not done at the end of the workflow execution, but are only updated periodically by the cronjob. 

How do you think we should handle these workflows? I think it might be confusing for the user to see some workflows disappear when changing their order.

CHANGES.rst
```diff
@@ -6,6 +6,7 @@ Version 0.9.0 (UNRELEASED)
 
 - Adds support for Kubernetes networking/v1 API to interactive sessions.
 - Changes workflow list endpoint to add the possibility to filter by workflow id.
+- Changes workflow list endpoint to add the possibility to sort by most disk and cpu quota used workflows.
```
mdonadoni, line 9:
```suggestion
- Changes workflow list endpoint to add the possibility to sort workflows by most disk and cpu quota used.
```

reana_workflow_controller/rest/workflows.py
audrium: Thanks! Very nice catch, all my workflows had quotas while I tested and I didn't notice this behavior. Fixed it to include all the workflows!

mdonadoni: Didn't know about `nullslast`, nice!

PR #455, opened 2022-06-15
rest: change default mime type for files download

mdonadoni:
Changes the default MIME type used when downloading files to `application/octet-stream`.
Previously, `multipart/form-data` was used but the response was not correctly encoded as such.

Closes reanahub/reana-server#418

How to reproduce the issue:
1. Launch `reana-demo-helloworld`
2. Preview the file `data/names.txt`
3. Check the logs of reana-server

Expected result: the warning is not present anymore in reana-server's logs

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -1382,7 +1382,11 @@
           }
         ],
         "produces": [
-          "multipart/form-data"
+          "application/octet-stream",
+          "application/json",
+          "application/zip",
+          "image/*",
+          "text/html"
         ],
         "responses": {
           "200": {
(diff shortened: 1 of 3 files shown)


VMois (commented):
Warning disappeared.

*minor:* Looks like the return "Content-type" is always "application/octet-stream" for text, python, and JSON in web UI. Images and HTML files are fine. Not sure whether the client receives the same content type as web UI. It should use the same endpoint for downloading files.
 
Maybe, we should remove "application/json" from the OpenAPI spec if the JSON content-type is never returned.

VMois:
> However, in these cases (Python/JSON/text files) it might be better to use text/plain. Should we open an issue about this?

Let's open an issue not to lose track of it. I am not sure how important it is, but it doesn't break stuff so let's have an issue and deal with it later.

VMois (approved):
PR solves the issue. Nice work!

Let's create a minor issue to investigate/fix encoding type problems for text, py, and JSON files.

PR #461, opened 2022-07-12
return error if workflow is deleted without workspace

### LINKED ISSUE #460: rest: workspace needs to be deleted when deleting a workflow
Following https://github.com/reanahub/reana-client/issues/614 and https://github.com/reanahub/reana-ui/issues/280, we should return an error when a request tries to delete a workflow without also deleting its workspace. 

https://github.com/reanahub/reana-workflow-controller/blob/1993a858a537c37c77626429b4624850d2d1e159/reana_workflow_controller/rest/workflows_status.py#L523-L526

VMois:
closes #460

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,7 @@ Version 0.9.0 (UNRELEASED)
 - Changes the workflow status endpoint to update the disk quota usage when a workspace is deleted.
 - Changes the CWD of jupyter's terminals to the directory of the workflow's workspace.
 - Changes the k8s specification of interactive sessions' pods to remove the environment variables used for service discovery.
+- Changes the workflow set status endpoint to remove the workspace by default when a workflow is deleted and to fail if the request is asking not to delete the workspace.
 
 Version 0.8.1 (2022-02-07)
 ---------------------------
--- a/reana_workflow_controller/rest/workflows_status.py
+++ b/reana_workflow_controller/rest/workflows_status.py
@@ -522,7 +522,16 @@ def set_workflow_status(workflow_id_or_name):  # noqa
             )
         elif status == DELETED:
             all_runs = True if request.json.get("all_runs") else False
-            workspace = True if request.json.get("workspace") else False
+            workspace = True if request.json.get("workspace", True) else False
+            if not workspace:
+                return (
+                    jsonify(
+                        {
+                            "message": "Workspace must always be deleted when deleting a workflow.",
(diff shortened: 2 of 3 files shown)


reana_workflow_controller/rest/workflows_status.py
```diff
@@ -522,7 +522,16 @@ def set_workflow_status(workflow_id_or_name):  # noqa
             )
         elif status == DELETED:
             all_runs = True if request.json.get("all_runs") else False
-            workspace = True if request.json.get("workspace") else False
+            workspace = True if request.json.get("workspace", True) else False
```
mdonadoni, line 525: I think we can leave this line as it was before, so that we don't change the default value of `workspace`. WDYT?

reana_workflow_controller/rest/workflows_status.py
```diff
@@ -522,7 +522,16 @@ def set_workflow_status(workflow_id_or_name):  # noqa
...
+            if not workspace:
+                return (
+                    jsonify(
+                        {
+                            "message": "Workspace must be deleted with the workflow. Use workspace=True.",
```
mdonadoni, line 530:
```suggestion
                            "message": "Workspace must always be deleted when deleting a workflow.",
```
Given that this message will be shown to users having an old version of `reana-client`, what do you think about changing it to remove any reference to the `workspace` parameter?

mdonadoni (approved):
Works well! :smile:

reana_workflow_controller/rest/workflows_status.py
VMois: In this case, when `workspace` is not specified, it will become `False` and fail.

CHANGES.rst
```diff
@@ -15,6 +15,7 @@ Version 0.9.0 (UNRELEASED)
 - Changes the workflow status endpoint to update the disk quota usage when a workspace is deleted.
 - Changes the CWD of jupyter's terminals to the directory of the workflow's workspace.
 - Changes the k8s specification of interactive sessions' pods to remove the environment variables used for service discovery.
+- Changes workflow set status endpoint when workflow is deleted to remove workspace by default and fail if request is asking not to delete workspace.
```
mdonadoni, line 18:
```suggestion
- Changes the workflow set status endpoint to remove the workspace by default when a workflow is deleted and to fail if the request is asking not to delete the workspace.
```
WDYT?

PR #463, opened 2022-07-27
rest: allow starting workflows without retention rules

mdonadoni:
After reanahub/reana-server#528, it can happen that workflows do not
have any retention rules.

Closes reanahub/reana#653


VMois (approved):
Nice catch!

PR #465, opened 2022-08-05
rest: correctly parse boolean query parameters in `get_workflows`

### LINKED ISSUE #464: rest: boolean query parameters are not parsed correctly
Some of the boolean query parameters are not correctly parsed, for example:

https://github.com/reanahub/reana-workflow-controller/blob/5eb424ffbcc54be68fae0f6845e7fec539020ccd/reana_workflow_controller/rest/workflows.py#L242-L244

In particular, `request.args.get` returns a string that needs to be converted to a boolean value. If this conversion is not done and one of the parameters is set to `false`, its value will be `"false"` (string) instead of `False` (bool). However, `"false"` is a thruthy value in Python, given that it is a non-empty string.

mdonadoni:
Closes #464

### DIFF
--- a/reana_workflow_controller/factory.py
+++ b/reana_workflow_controller/factory.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2017, 2018, 2019, 2020, 2021 CERN.
+# Copyright (C) 2017, 2018, 2019, 2020, 2021, 2022 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -12,11 +12,34 @@ from __future__ import absolute_import
 
 import logging
 
-from flask import Flask
+from flask import Flask, jsonify
+from marshmallow.exceptions import ValidationError
 from reana_commons.config import REANA_LOG_FORMAT, REANA_LOG_LEVEL
 from reana_db.database import Session
+from werkzeug.exceptions import UnprocessableEntity
+
 
 from reana_db.models import Base  # isort:skip  # noqa
+
+
+def handle_args_validation_error(error: UnprocessableEntity):
+    """Error handler for werkzeug exception ``UnprocessableEntity``.
+
+    This error handler is needed to display useful error messages, instead of the
+    generic default one, when marshmallow argument validation fails.
+    """
+    error_message = error.description or str(error)
+
+    exception = getattr(error, "exc", None)
+    if isinstance(exception, ValidationError):
+        validation_messages = []
+        for field, messages in exception.normalized_messages().items():
+            validation_messages.append(
+                "Field '{}': {}".format(field, ", ".join(messages))
+            )
+        error_message = ". ".join(validation_messages)
+
+    return jsonify({"message": error_message}), 400
 
 
 def create_app(config_mapping=None):
@@ -40,5 +63,8 @@ def create_app(config_mapping=None):
     app.register_blueprint(workflows.blueprint, url_prefix="/api")
     app.register_blueprint(workflows_status.blueprint, url_prefix="/api")
     app.register_blueprint(workflows_workspace.blueprint, url_prefix="/api")
+
+    app.register_error_handler(UnprocessableEntity, handle_args_validation_error)
+
     app.session = Session
     return app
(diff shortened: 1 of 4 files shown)


reana_workflow_controller/factory.py
```diff
@@ -12,13 +12,36 @@
...
+from marshmallow.exceptions import ValidationError
 from reana_commons.config import REANA_LOG_FORMAT, REANA_LOG_LEVEL
 from reana_db.database import Session
+from werkzeug.exceptions import UnprocessableEntity
+
 
 from reana_db.models import Base  # isort:skip  # noqa
 
 
+def handle_args_validation_error(error: UnprocessableEntity):
```
mdonadoni, line 25: This is the same error handler we use in `reana-server`. Should I move this to `reana-commons`?

VMois:
*suggestion:* I would change `use_kwargs` to `use_args` to reduce the number of arguments in the function and have better control over them (check `verbose` use case). Also, if we will use `Schema` in future, `use_kwargs` cannot be used ([link](https://webargs.readthedocs.io/en/latest/advanced.html#when-to-avoid-use-kwargs)).

I have also used the `missing` keyword to set default values in case they are missing. It is nice to keep them in one place.

I think we should use the `missing` keyword. What do you think about `use_args`?

```diff
diff --git a/reana_workflow_controller/rest/workflows.py b/reana_workflow_controller/rest/workflows.py
index 329e5ac..e20cbd1 100644
--- a/reana_workflow_controller/rest/workflows.py
+++ b/reana_workflow_controller/rest/workflows.py
@@ -9,6 +9,7 @@
 """REANA Workflow Controller workflows REST API."""

 import json
+from typing import Optional
 from uuid import uuid4

 from flask import Blueprint, jsonify, request
@@ -19,7 +20,7 @@ from reana_db.models import User, Workflow, RunStatus, WorkflowResource
 from reana_db.utils import _get_workflow_with_uuid_or_name, get_default_quota_resource
 from sqlalchemy import and_, nullslast
 from webargs import fields
-from webargs.flaskparser import use_kwargs
+from webargs.flaskparser import use_args


 from reana_workflow_controller.config import DEFAULT_NAME_FOR_WORKFLOWS
@@ -48,32 +49,23 @@ blueprint = Blueprint("workflows", __name__)

 @blueprint.route("/workflows", methods=["GET"])
 @use_paginate_args()
-@use_kwargs(
+@use_args(
     {
         "include_progress": fields.Bool(),
         "include_workspace_size": fields.Bool(),
         "include_retention_rules": fields.Bool(),
-        "search": fields.String(),
-        "sort": fields.String(),
-        "status": fields.String(attribute="status_list"),
-        "type": fields.String(attribute="type_", required=True),
+        "search": fields.String(missing=""),
+        "sort": fields.String(missing="desc"),
+        "status": fields.String(missing=""),
+        "type": fields.String(required=True),
         "user": fields.String(attribute="user_uuid", required=True),
-        "verbose": fields.Bool(),
+        "verbose": fields.Bool(missing=False),
         "workflow_id_or_name": fields.String(),
     },
     location="query",
 )
 def get_workflows(  # noqa
-    include_progress=None,
-    include_workspace_size=None,
-    include_retention_rules=None,
-    search="",
-    sort="desc",
-    status_list="",
-    type_="batch",
-    user_uuid=None,
-    verbose=False,
-    workflow_id_or_name=None,
+    args,
     paginate=None,
 ):
     r"""Get all workflows.
@@ -260,14 +252,18 @@ def get_workflows(  # noqa
                 "message": "Internal workflow controller error."
               }
     """
-    try:
-        if include_progress is None:
-            include_progress = verbose
-        if include_workspace_size is None:
-            include_workspace_size = verbose
-        if include_retention_rules is None:
-            include_retention_rules = verbose
+    user_uuid: str = args["user_uuid"]
+    type_: str = args["type"]
+    verbose: bool = args["verbose"]
+    sort: str = args["sort"]
+    search: str = args["search"]
+    status_list: str = args["status"]
+    include_progress: bool = args.get("include_progress", verbose)
+    include_workspace_size: bool = args.get("include_workspace_size", verbose)
+    include_retention_rules: bool = args.get("include_retention_rules", verbose)
+    workflow_id_or_name: Optional[str] = args.get("workflow_id_or_name")

+    try:
         user = User.query.filter(User.id_ == user_uuid).first()
         if not user:
             return jsonify({"message": "User {} does not exist".format(user_uuid)}), 404
```

On a side note, whatever approach we choose, we can use the code from this PR as a reference for other places.

reana_workflow_controller/factory.py
VMois: I am a bit biased here because I don't like `r-commons` and would suggest not to put it there. But another reason is that `webargs` versions differ between r-w-c and r-server. 

If you need someone else's opinion on this, feel free to tag them here.

mdonadoni: It mostly depends on whether we will continue using `webargs`, whether we will be able to update it to have the same version everywhere and if we will use it also in other components like r-job-controller. Maybe I can just put a `# FIXME` comment for now?

VMois (approved):

PR #494, opened 2023-01-12
workflow_run_manager: do not mount db directory in job-controller

mdonadoni:
Closes reanahub/reana#686

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,7 @@ Version 0.9.0 (UNRELEASED)
 - Changes the workflow set status endpoint to remove the workspace by default when a workflow is deleted and to fail if the request is asking not to delete the workspace.
 - Changes to PostgreSQL 12.13.
 - Changes the `move_files` endpoint to allow moving files while a workflow is running.
+- Changes the deployment of job-controller to avoid unnecessarily mounting the database's directory.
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -34,7 +34,6 @@
         "sphinxcontrib-httpdomain>=1.5.0",
         "sphinxcontrib-openapi>=0.3.0",
         "sphinxcontrib-redoc>=1.5.1",
-        "mistune<2.0.0",
```
tiborsimko, line 37: Perhaps we should also upgrade the lower version of `sphinxcontrib-openapi` to 0.8.0 or higher, so that it'll be fully clear that the mistune pin can be removed.

tiborsimko: Also, it would be good to invert the two commits, i.e. having mistune fix come first and the volume mounting second, so that both commits of the branch would pass the CI.

tiborsimko (approved):
Works well. Left some cosmetic comments. Perhaps you can leave a note in CHANGES file about this change. Although nobody should really need the old behaviour.

setup.py
tiborsimko: Strictly speaking, it is only necessary there where we pinned `mistune` dependency, i.e. only here. Because the other repositories are working nicely with older versions too.

That said, it may be nice indeed to update others, for consistency's sake, since we are touching all the repositories?

Issue #498, opened 2023-02-02
configurable auto-closure of interactive sessions after days of inactivity


tiborsimko:
**Current behaviour**

When a user opens an interactive session upon a workflow, the sessions stays running "forever", unless they are explicitly closed by the user.  This can lead to their unnecessary accumulation over time, because notebooks are used only to "accompany" wokflows.

Note that we are improving the behaviour to automatically close any notebooks when a user deletes a  workflow, see: https://github.com/reanahub/reana-client/issues/649

However, it would be even better to foresee an automatic closure of all sessions that run for longer than some amount of hours or days  without any user activity.

**Expected behaviour**

We can introduce a new configurable option that would allow cluster administrator to set a value during Helm cluster deployment that would instruct REANA to automatically close any open interactive sessions if they have been inactive for N days.

See for example https://github.com/jupyterhub/jupyterhub-idle-culler 

We should also make it very clear in the web UI, when a user opens a session, what its inactivity lifetime will be.

PR #500, opened 2023-02-08
k8s: update interactive sessions spec to include labels

audrium:
Add new labels for workflow_id, owner_id and workflow_mode for improved subset selection of objects and metadata parsing from the pod specification.

closes https://github.com/reanahub/reana-workflow-controller/issues/498

### DIFF
--- a/reana_workflow_controller/k8s.py
+++ b/reana_workflow_controller/k8s.py
@@ -38,11 +38,24 @@ class InteractiveDeploymentK8sBuilder(object):
        port exposed by deployment itself, the one which can change
        from one interactive session application to other."""
 
-    def __init__(self, deployment_name, workspace, image, port, path, cvmfs_repos=None):
+    def __init__(
+        self,
+        deployment_name,
+        workflow_id,
+        owner_id,
+        workspace,
+        image,
+        port,
+        path,
+        cvmfs_repos=None,
+    ):
         """Initialise basic interactive deployment builder for Kubernetes.
 
         :param deployment_name: Name which identifies all deployment objects
             and maps to the workflow it belongs.
+        :param workflow_id: UUID of the workflow to which the interactive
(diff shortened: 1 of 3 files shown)


reana_workflow_controller/k8s.py
```diff
@@ -137,12 +152,18 @@ def _build_deployment(self, metadata):
             enable_service_links=False,
             automount_service_account_token=False,
         )
+        labels = {
+            "app": self.deployment_name,
+            "reana_workflow_mode": "session",
+            "reana-run-session-workflow-uuid": str(self.workflow_id),
+            "reana-run-session-owner-id": str(self.owner_id),
```
mdonadoni, line 159:
```suggestion
            "reana-run-session-owner-uuid": str(self.owner_id),
```
WDYT? To be coherent with `reana-run-session-workflow-uuid`

mdonadoni, line 159: What do you think about standardizing labels name? I think we should decide whether we want to use `_` or `-` and whether we want to use the prefix `reana.io/...` (as we do for example for the `reana.io/system` label). This can be done in another PR.

reana_workflow_controller/k8s.py
```diff
@@ -252,6 +273,8 @@ def build_interactive_jupyter_deployment_k8s_objects(
     access_path,
     access_token=None,
     cvmfs_repos=None,
+    owner_id=None,
+    workflow_id=None,
```
mdonadoni, line 277: We need to update the docstring of this method to include the new parameters

reana_workflow_controller/k8s.py
audrium: Created an issue for this: https://github.com/reanahub/reana-workflow-controller/issues/503

mdonadoni (approved):

PR #502, opened 2023-02-16
k8s: update interactive sessions ingress with annotations

audrium:


reana_workflow_controller/k8s.py
```diff
@@ -109,7 +110,10 @@ def _build_ingress(self, metadata):
             api_version="networking.k8s.io/v1",
             kind="Ingress",
             spec=spec,
-            metadata=metadata,
+            metadata=client.V1ObjectMeta(
+                name=self.deployment_name,
+                annotations=REANA_INGRESS_ANNOTATIONS,
+            ),
```
mdonadoni, line 116: This discards the metadata passed to the function. What do you think about merging the ingress annotations with the metadata passed to the function?

audrium: It's not great to mutate the same object which is used elsewhere (annotations would be passed to `service` and `deployment` specs). We could create a deep copy of course, but I didn't want to pass additional info like `labels`  to the ingress spec from the metadata object, that's why I've created another one

mdonadoni: If it's not needed, let's remove it from the parameters of the function

mdonadoni (approved):

PR #514, opened 2023-07-14
consumer: rollback database transaction in case of errors

mdonadoni:
Rollback the database `Session` when an error occurs while handling a
status update message.

This fixes an issue where, if the current transaction fails for any
reason, then the consumer keeps consuming the queue without actually
updating the status of workflows and deleting the related `run-batch`
pods. This is caused by the fact that messages are acknowledged, but
querying the database fails, as the transaction was not rolled back.
This causes REANA to stop scheduling workflows, as finished workflows
are still considered as running.

### DIFF
--- a/reana_workflow_controller/consumer.py
+++ b/reana_workflow_controller/consumer.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2018, 2019, 2020, 2021, 2022 CERN.
+# Copyright (C) 2018, 2019, 2020, 2021, 2022, 2023 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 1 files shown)


reana_workflow_controller/consumer.py
```diff
@@ -126,14 +126,14 @@ def on_message(self, body, message):
                     f"Event for workflow {workflow_uuid} that doesn't exist in DB received:\n"
                     f"{body}\nIgnoring..."
                 )
-        except REANAWorkflowControllerError as rwce:
-            logging.error(rwce, exc_info=True)
```
mdonadoni, line 130: As far as I can tell, this exception cannot occur because it is raised by `_get_workflow_engine_pod_logs` and `_delete_workflow_job`, but both of them have surrounding `try ... catch`es when called that catch this exception type.

giuseppe-steduto (commented):
Looks good! Just a quick question: what would happen if the exception was raised due to a connection error to the DB? Would `Session.rollback()` then raise another (uncaught) exception?

P.S. Needs a rebase!

mdonadoni:
> Looks good! Just a quick question: what would happen if the exception was raised due to a connection error to the DB? Would `Session.rollback()` then raise another (uncaught) exception?

It shouldn't, as a quick test:
```python
>>> from ... import ...
>>> test_id = "..."
>>>
>>> wf = Workflow.query.filter_by(id_=test_id).one()
>>> wf.status = "failed"
[... stopping the database now ...]
>>> Session.commit()
[...]
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
[...]
>>> Session.rollback()
[... all good ...]
```

>  P.S. Needs a rebase!

Done!

giuseppe-steduto (approved):

PR #518, opened 2023-07-20
global: update to apispec v3

mdonadoni:
Update to apispec v3, as the version used before is not compatible with
PyYAML v6.

Closes reanahub/reana-commons#399


mdonadoni:
CI should pass after releasing new alpha version of reana-commons and pytest-reana

giuseppe-steduto (approved):

PR #532, opened 2023-10-25
utils: return different message when deleting all the workflow runs

### LINKED ISSUE #531: rest: change returned message when deleting multiple workflows
> Very minor: maybe the text of the success notification could be different when deleting _all the runs_ of a workflow (like in `r-client`)? Could be something as simple as changing "workflow" to "workflows" in "workflow successfully deleted", or could be a totally different sentence, like `r-client` does. 

> Note that if we want to address this, rather than changing `r-ui`, it might be better to change the `message` property returned by `r-w-c` when deleting more than one run.

_Originally posted by @giuseppe-steduto in https://github.com/reanahub/reana-ui/pull/357#discussion_r1360689819_

giuseppe-steduto:
Change the `delete_workflow` function so that when the deletion of a
workflow includes all its runs, then this is mentioned in the `message`
property returned.

Closes #531.

### DIFF
--- a/reana_workflow_controller/rest/utils.py
+++ b/reana_workflow_controller/rest/utils.py
@@ -270,10 +270,16 @@ def delete_workflow(workflow, all_runs=False, workspace=False):
                 _mark_workflow_as_deleted_in_db(workflow)
                 remove_workflow_jobs_from_cache(workflow)
 
+            if all_runs:
+                message = "All workflows named {0} successfully deleted.".format(
+                    workflow.name
+                )
+            else:
+                message = "Workflow successfully deleted."
             return (
                 jsonify(
                     {
-                        "message": "Workflow successfully deleted",
+                        "message": message,
                         "workflow_id": workflow.id_,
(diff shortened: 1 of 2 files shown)


reana_workflow_controller/rest/utils.py
```diff
@@ -270,10 +270,16 @@ def delete_workflow(workflow, all_runs=False, workspace=False):
                 _mark_workflow_as_deleted_in_db(workflow)
                 remove_workflow_jobs_from_cache(workflow)
 
+            if all_runs:
+                message = "All workflows named {0} successfully deleted.".format(
+                    workflow.name
+                )
```
mdonadoni, line 276: minor: just as a note, this could be rewritten using f-strings, which tend to be more compact:
```python
message = f"All workflows named {workflow.name} successfully deleted."
```
(I am going to merge anyway, no need to change this!)

mdonadoni (approved):
LGTM!

PR #534, opened 2023-10-27
ci: publish docker image after new release

mdonadoni:
Closes reanahub/reana#752


tiborsimko (approved):
Cosmetics: let's capitalise the "name" as in "Release Docker image" for consistency with the other rules we have, and let's order the "needs" alphabetically.

PR #540, opened 2023-11-09
k8s: allow automounting of CVMFS repositories

mdonadoni:
Closes reanahub/reana-commons#423


mdonadoni:
CI should pass after releasing new alpha version of r-commons and bumping version in setup.py

PR #542, opened 2023-11-14
rest: add get_workflow_share_status endpoint

DaanRosendal:
Adds a new endpoint to retrieve whom a workflow is shared with.

Closes reanahub/reana-client#686

### DIFF
--- a/reana_workflow_controller/rest/workflows.py
+++ b/reana_workflow_controller/rest/workflows.py
@@ -1146,3 +1146,219 @@ def share_workflow(
     except Exception as e:
         logging.exception(str(e))
         return jsonify({"message": str(e)}), 500
+
+
+@blueprint.route("/workflows/<workflow_id_or_name>/unshare", methods=["POST"])
+@use_kwargs(
+    {
+        "user_id": fields.Str(required=True),
+        "user_email_to_unshare_with": fields.Str(required=True),
+    },
+    location="query",
+)
+def unshare_workflow(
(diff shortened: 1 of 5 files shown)


reana_workflow_controller/rest/workflows.py
```diff
@@ -1362,3 +1362,155 @@ def unshare_workflow(
...
+                "workflow_id": "256b25f4-4cfb-4684-b7a8-73872ef455a1",
+                "workflow_name": "mytest.1",
+                "shared_with": [
+                    {
+                      "user_email": "bob@example.org",
+                      "valid_until": "2022-11-24T23:59:59"
+                    }
+                ]
+              }
+        401:
```
mdonadoni, line 1432: I think the 401 and 403 errors are never returned? They are normally returned by reana-server, but not by reana-workflow-controller

reana_workflow_controller/rest/workflows.py
```diff
@@ -1362,3 +1362,155 @@ def unshare_workflow(
     except Exception as e:
         logging.exception(str(e))
         return jsonify({"message": str(e)}), 500
+
+
+@blueprint.route("/workflows/<workflow_id_or_name>/share-status", methods=["GET"])
```
mdonadoni, line 1367: Let's change `share-status` to `share_status`, to be consistent with the rest (e.g. we already have `move_files`, `retention_rules`, `disk_usage`, ...)

Issue #546, opened 2023-11-22
job pods keep running even after workflow failure


mdonadoni:
The pod `reana-run-batch-...` is terminated as soon as one job of the workflow fails, even if  there are more jobs running:
https://github.com/reanahub/reana-workflow-controller/blob/3004b14a7d60eb39dfcbdc51e15242b27edd70c3/reana_workflow_controller/consumer.py#L163-L170

This means that those jobs will outlive `reana-run-batch-...`, the k8s pod are not  cleaned up and the database is not updated.

How to reproduce:

1. Prepare workflow files (see below)
2. Run the workflow

One job will fail, the other one will continue running even after `reana-run-batch-...` is terminated.
The job pod will not be cleaned up either.

<details>
<summary>values-dev.yaml</summary>

```diff
         REANA_RATELIMIT_SLOW: "5 per second"
     reana_workflow_controller:
       image: docker.io/reanahub/reana-workflow-controller
-      environment:
-        REANA_RUNTIME_KUBERNETES_KEEP_ALIVE_JOBS_WITH_STATUSES: failed
+      # environment:
+      #   REANA_RUNTIME_KUBERNETES_KEEP_ALIVE_JOBS_WITH_STATUSES: failed
     reana_workflow_engine_cwl:
       image: docker.io/reanahub/reana-workflow-engine-cwl
     reana_workflow_engine_yadage:
```

</details>

<details>
<summary>reana.yaml</summary>

```yaml
version: 0.9.0
inputs:
  files:
    - Snakefile
workflow:
  type: snakemake
  file: Snakefile
```
</details>

<details>
<summary>Snakefile</summary>

```snakefile
rule all:
    input:
        "r1.txt",
        "r2.txt",

rule r1:
    output:
        "r1.txt"
    container:
        "docker://docker.io/library/python:3.8-slim"
    shell:
        "sleep 120; echo done > r1.txt"

rule r2:
    output:
        "r2.txt"
    container:
        "docker://docker.io/library/python:3.8-slim"
    shell:
        "exit 1"
```
</details>

Issue #252, opened 2019-09-03
refactor: stop batch workflow run to stop only workflow engine pod


diegodelemos:
### Why is it like this now?
When we re-implemented the workflow stop feature we took a shortcut due time constraints, implementing the whole stop feature inside workflow controller, including workflow engine and all running jobs.

https://github.com/reanahub/reana-workflow-controller/blob/2b47b4d5d5b29a87e2ba8a222b5a9daeb6166993/reana_workflow_controller/workflow_run_manager.py#L296-L309

This works for Kubernetes because [we already depend on Kubernetes Python library](https://github.com/reanahub/reana-workflow-controller/blob/2b47b4d5d5b29a87e2ba8a222b5a9daeb6166993/setup.py#L54) because of the [`KubernetesWorkflowRunManager`](https://github.com/reanahub/reana-workflow-controller/blob/2b47b4d5d5b29a87e2ba8a222b5a9daeb6166993/reana_workflow_controller/workflow_run_manager.py#L185) implementation.

### Why it cannot continue like this?

1. Stop won't work for any other compute backend than Kubernetes
2. This is not the desired final state, since the code that stops the Jobs should be abstracted away, and use the corresponding library to connect compute backend depending of the job's backend type. If we would do that we would mix RWC and RJC responsibilities.

   [RJC already installs the bindings to connect to the underlying compute backends](https://github.com/reanahub/reana-job-controller/blob/47a411cfc6fa756894c9dd91483547f2b7586335/setup.py#L51-L54) (until now Kubernetes and HTCondor). And it offers a [REST API to delete jobs](https://reana-job-controller.readthedocs.io/en/latest/_static/api.html#operation/delete_job) created in the underlying compute backend, and it's [implementation is already backend aware](https://github.com/reanahub/reana-job-controller/blob/47a411cfc6fa756894c9dd91483547f2b7586335/reana_job_controller/rest.py#L380-L381) for example, [this would be the logic to stop in HTCondor](https://github.com/reanahub/reana-job-controller/blob/47a411cfc6fa756894c9dd91483547f2b7586335/reana_job_controller/htcondorcern_job_manager.py#L230) and [this would be in Kubernetes](https://github.com/reanahub/reana-job-controller/blob/47a411cfc6fa756894c9dd91483547f2b7586335/reana_job_controller/rest.py#L380-L381).

### Solution

[...]

PR #559, opened 2024-01-23
fix runaway pods

mdonadoni:
- refactor(workflow-run-manager): do not update status of jobs (#559)
- feat(run-manager): call shutdown endpoint before workflow stop (#559)
- fix(run-manager): graceful shutdown of job-controller (#559)

Closes https://github.com/reanahub/reana-workflow-controller/issues/252
Closes https://github.com/reanahub/reana-workflow-controller/issues/546
Closes https://github.com/reanahub/reana-job-controller/issues/347


giuseppe-steduto (approved):
Nice and clean approach! I was wondering whether we should increase the 
termination grace period to be more sure that the HTTP call is completed, but I guess what really matters is that the request is sent to `/shutdown`. Then even if the response is not delivered in time to the `run-batch` pod (because the grace period has passed), the jobs should have terminated anyway.

As for the commit scopes, maybe it's better to always use `run-manager` rather than `workflow-run-manager`?

PR #563, opened 2024-01-29
fix(stop): store engine logs of stopped workflow (#563)

### LINKED ISSUE #560: stopped workflow do not have engine logs
If a workflow is stopped, then the workflow engine logs are not fetched and stored in the database

mdonadoni:
Closes #560

Depends on #559

### DIFF
--- a/reana_workflow_controller/rest/utils.py
+++ b/reana_workflow_controller/rest/utils.py
@@ -64,6 +64,7 @@ from reana_workflow_controller.config import (
     REANA_GITLAB_HOST,
     PREVIEWABLE_MIME_TYPE_PREFIXES,
 )
+from reana_workflow_controller.consumer import _update_workflow_status
 from reana_workflow_controller.errors import (
     REANAExternalCallError,
     REANAWorkflowControllerError,
@@ -135,11 +136,8 @@ def start_workflow(workflow, parameters):
 
 def stop_workflow(workflow):
     """Stop a given workflow."""
-    if workflow.status == RunStatus.running:
-        kwrm = KubernetesWorkflowRunManager(workflow)
-        kwrm.stop_batch_workflow_run()
-        workflow.status = RunStatus.stopped
-        Session.add(workflow)
+    if workflow.can_transition_to(RunStatus.stopped):
+        _update_workflow_status(workflow, RunStatus.stopped, logs="")
         Session.commit()
     else:
         message = ("Workflow {id_} is not running.").format(id_=workflow.id_)
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -823,12 +823,12 @@ def test_start_already_started_workflow(
 
 @pytest.mark.parametrize(
     "current_status, expected_status, expected_http_status_code, "
-    "k8s_stop_call_count",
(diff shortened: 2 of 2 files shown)


mdonadoni:
Note that the method `_update_workflow_status` in `consumer.py` is now being called in `rest/utils.py`. I think a refactor is needed, in particular:
1. The logic on how to update the status/logs/job progress/job cache of workflows should not be in `consumer.py`, but in a dedicated module/package
2. There should only be one way (thus, one function) to stop a workflow (and to start/delete/... it), which both updates the database and interacts with k8s as needed. These functions would/should use the ones from (1), and should not be part of `rest`.

Proper design/planning is needed, though. Should we do it later or as part of this PR?

reana_workflow_controller/rest/utils.py
```diff
@@ -135,11 +136,8 @@ def _start_workflow_db(workflow, parameters):
...
 def stop_workflow(workflow):
     """Stop a given workflow."""
-    if workflow.status == RunStatus.running:
-        kwrm = KubernetesWorkflowRunManager(workflow)
-        kwrm.stop_batch_workflow_run()
-        workflow.status = RunStatus.stopped
-        Session.add(workflow)
+    if workflow.can_transition_to(RunStatus.stopped):
+        _update_workflow_status(workflow, RunStatus.stopped, logs="")
         Session.commit()
```
giuseppe-steduto, line 141: I guess calling `Session.commit()` here is unnecessary, since it is already being called by [`_update_workflow_status`](https://github.com/mdonadoni/reana-workflow-controller/blob/e1f8b39f37e528ea762f7caa24a275927d58aee7/reana_workflow_controller/consumer.py#L143) ([here](https://github.com/reanahub/reana-db/blob/cc6052242fd14cf3413b793d0aa32a24871fe1b1/reana_db/models.py#L755) the `reana_db` code), right?

giuseppe-steduto (commented):
I fully support the idea of a refactor, especially to create a single point from which to update the status of a workflow. However, maybe it's better if we create a ticket for it and work on it later, as part of the next minor release, so we can focus on finishing what's left for the patch release and fixing this bug without the risk of changing the code in a lot of places. WDYT?

reana_workflow_controller/rest/utils.py
mdonadoni: It is still needed because even though the status is commited by `Workflow.update_workflow_status`, the logs are fetched and saved to the database after that:
https://github.com/reanahub/reana-workflow-controller/blob/817b019b3745862436e99570c10c6d8ea35533f4/reana_workflow_controller/consumer.py#L143-L156

mdonadoni:
> I fully support the idea of a refactor, especially to create a single point from which to update the status of a workflow. However, maybe it's better if we create a ticket for it and work on it later, as part of the next minor release, so we can focus on finishing what's left for the patch release and fixing this bug without the risk of changing the code in a lot of places. WDYT?

I agree! I have opened a new issue: https://github.com/reanahub/reana-workflow-controller/issues/565

giuseppe-steduto (approved):

PR #571, opened 2024-02-22
feat(manager): pass custom env variables to workflow engines and job controller (#571)

mdonadoni:
- feat(manager): pass custom env variables to workflow engines (#571)
- feat(manager): pass custom env variables to job controller (#571)

### DIFF
--- a/reana_workflow_controller/config.py
+++ b/reana_workflow_controller/config.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2017, 2018, 2019, 2020, 2021, 2022, 2023 CERN.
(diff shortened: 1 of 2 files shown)


reana_workflow_controller/config.py
```diff
@@ -117,6 +117,31 @@
 ]
 """Common to all workflow engines environment variables."""
 
+
+def _env_vars_dict_to_k8s_list(env_vars):
+    return [{"name": name, "value": value} for name, value in env_vars.items()]
```
giuseppe-steduto, line 122: Maybe we could move this function to the top of the file, just to avoid having it in the middle of the variables declaration?

giuseppe-steduto (approved):
LGTM, just a very minor cosmetic comment (you can also update the copyright notice years)!

PR #573, opened 2024-02-26
docs(openapi): amend response description for file deletion (#573)

giuseppe-steduto:


docs/openapi.json
```diff
@@ -1453,7 +1453,7 @@
         ],
         "responses": {
           "200": {
-            "description": "Requests succeeded. The file has been downloaded.",
+            "description": "Requests succeeded. Details about deleted files and failed deletions are returned.",
```
tiborsimko, line 1456: Isn't it rather "Request succeeded" in singular? (as in other REANA components)

tiborsimko: Amended.

tiborsimko (approved):

PR #582, opened 2024-04-17
feat(sessions): add allowlist for interactive session images (#582)

### LINKED ISSUE #569: interactive-sessions: implement configurable allowlist
Let's add a configurable allowlist of docker images among which users can choose when starting a new interactive session.

Things to do:
- [ ] Decide how to let the admins customise this allowlist (most probably helm value?)
- [ ] Implement the actual allowlist checks in reana-workflow-controller
- [ ] Decide whether to let users choose the docker image also from the UI. In this case, the list needs to be exposed also in web interface.
- [ ] Bonus point: instead of a simple allowlist, consider making customisable also other details about the docker images (e.g. web port used, env variables needfed, etc.). This is a much bigger task, though.

Note that not all docker images are compatible with the way we currently deploy jupyter notebooks (e.g. they might expose a different port, etc.)

mdonadoni:
Closes #569

### DIFF
--- a/reana_workflow_controller/config.py
+++ b/reana_workflow_controller/config.py
@@ -154,10 +154,60 @@ DEBUG_ENV_VARS = (
 )
 """Common to all workflow engines environment variables for debug mode."""
 
-JUPYTER_INTERACTIVE_SESSION_DEFAULT_IMAGE = (
-    "docker.io/jupyter/scipy-notebook:notebook-6.4.5"
+
+def _parse_interactive_sessions_environments(env_var):
+    config = {}
+    for type_ in env_var:
+        recommended = []
+        env_recommended = env_var[type_].get("recommended") or []
+        for recommended_item in env_recommended:
+            image = recommended_item.get("image")
+            if not image:
+                continue
+            name = recommended_item.get("name") or image
+            recommended.append({"name": name, "image": image})
+
+        config[type_] = {
+            "allow_custom": env_var[type_].get("allow_custom", False),
+            "recommended": recommended,
+        }
+    return config
+
+
+REANA_INTERACTIVE_SESSIONS_ENVIRONMENTS = _parse_interactive_sessions_environments(
+    json.loads(os.getenv("REANA_INTERACTIVE_SESSIONS_ENVIRONMENTS", "{}"))
 )
-"""Default image for Jupyter based interactive session deployments."""
+"""Allowed and recommended environments to be used for interactive sessions.
+
+This is a dictionary where keys are the type of the interactive session.
+For each session type, a list of recommended Docker images are provided (`recommended`)
+and whether custom images are allowed (`allow_custom`).
+
+Example:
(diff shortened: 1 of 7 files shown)


mdonadoni:
This works well also with reana-client, but there is currently no way to see which images are allowed directly from the CLI.
Should I implement a new command in reana-client, or do we want to use `reana-client info`?

Finally, the list of allowed images is now returned by `/config`, which is an endpoint that does not require the user to be authenticated, as it's supposed to return the configuration of the web UI. Do we want to create a new authenticated endpoint just to serve the configuration of interactive sessions? Or maybe should we use `/info`?
Note that `/config` seems to be only used in the UI, while `/info` only by reana-client.

reana_workflow_controller/config.py
```diff
@@ -154,10 +154,40 @@ def _env_vars_dict_to_k8s_list(env_vars):
...
+                "name": "Jupyter SciPy Notebook 6.4.5",
+                "image": "docker.io/jupyter/scipy-notebook:notebook-6.4.5"
+            }
+        ],
+        "allow_custom": true
+    }
+}
+"""
+
+REANA_INTERACTIVE_SESSIONS_RECOMMENDED_IMAGES = {
```
tiborsimko, line 180: Can we amend this to be ready for situations where the configuration is empty?

For example, some site may want to prevent opening of notebooks altogether by configuring:

```yaml
interactive_sessions:
  environments:
    jupyter:
      recommended:
      allow_custom: false
```

This currently leads to a traceback when starting reana-workflow-controller:

```
Traceback (most recent call last):
  File "/code/./reana_workflow_controller/app.py", line 17, in <module>
    app = create_app()
  File "/code/./reana_workflow_controller/factory.py", line 49, in create_app
    app.config.from_object("reana_workflow_controller.config")
  File "/usr/local/lib/python3.8/dist-packages/flask/config.py", line 227, in from_object
    obj = import_string(obj)
  File "/usr/local/lib/python3.8/dist-packages/werkzeug/utils.py", line 595, in import_string
    __import__(import_name)
  File "/code/./reana_workflow_controller/config.py", line 180, in <module>
    REANA_INTERACTIVE_SESSIONS_RECOMMENDED_IMAGES = {
  File "/code/./reana_workflow_controller/config.py", line 181, in <dictcomp>
    type_: {recommended["image"] for recommended in config["recommended"]}
TypeError: 'NoneType' object is not iterable
```

This means that the REANA would be basically down if people misconfigure the recommended images section. 

Can you make the YAML value detection more robust to prevent these cases? E.g. we could use something like `config.get("recommended", [])`  and such. We should also assume that  "image" part might be misconfigured.

mdonadoni: Done, it should now be more robust in case of missing recommended images, missing `allow_custom`, both here and in reana-ui

tiborsimko (approved):

Issue #587, opened 2024-06-13
config: allow customising ingress hosts


tiborsimko:
For the SSL-RIVER deployment, the opening of notebooks necessitates the creation of ingresses with a dedicated host. (Currently we use `*` which is not working.)

Basically we need to allow the following:

```console
$ kubectl get ingresses
NAME                                                     CLASS           HOSTS         ADDRESS   PORTS     AGE
reana-ingress                                            reana-traefik   *                       80, 443   9d
reana-run-session-e32d22d2-35eb-41c9-86ae-cd39ef39596d   reana-traefik   foo.bar.com             80        5d21h
```

The above test was performed by:

```console
$ kubectl edit ingress reana-run-session-e32d22d2-35eb-41c9-86ae-cd39ef39596d 
$ kubectl get ingress reana-run-session-e32d22d2-35eb-41c9-86ae-cd39ef39596d -o yaml
...
spec:
  ingressClassName: reana-traefik
  rules:
  - host: foo.bar.com
    http:
      paths:
      - backend:
```

The goal of this issue is to introduce a new configuration variable `REANA_INGRESS_HOST`, alongside existing variables `REANA_INGRESS_ANNOTATIONS` and `REANA_INGRESS_CLASS_NAME`, that would allow to customize the `host` part of the ingress rules.

PR #588, opened 2024-06-18
feat(helm): allow cluster administrator to configure ingress host (#588)

Alputer:
Closes reanahub/reana-workflow-controller#587


tiborsimko (approved):
Works nicely locally and on DEV :+1:

PR #591, opened 2024-06-26
feat(sessions): expose user secrets in interactive sessions (#591)

mdonadoni:
Closes reanahub/reana#805


tiborsimko (approved):
Works nicely :+1:

PR #594, opened 2024-08-06
fix(manager): reduce size of job-controller's db connection pool (#594)

mdonadoni:
Closes reanahub/reana#819


tiborsimko (approved):
Works nicely :+1: but needs a rebase. 

BTW `fix(manager)` or `perf(manager)`?  (Even though this change is not strictly speaking optimising performance by itself, it optimises the overall cluster performance...)

PR #595, opened 2024-08-07
fix(delete_workflow): set retention rules as inactive (#595)

mdonadoni:
Closes reanahub/reana-server#693

### DIFF
--- a/reana_workflow_controller/rest/utils.py
+++ b/reana_workflow_controller/rest/utils.py
@@ -234,14 +234,16 @@ def delete_workflow(workflow, all_runs=False, workspace=False):
                     .all()
                 )
             for workflow in to_be_deleted:
+                # 1. Stop open interactive sessions
                 int_session = workflow.sessions.first()
                 if int_session:
                     kwrm = KubernetesWorkflowRunManager(workflow)
                     kwrm.stop_interactive_session(int_session.id_)
 
                 if workspace:
+                    # 2. delete the workspace
                     remove_workflow_workspace(workflow.workspace_path)
-
+                    # 3. update the disk usage of the user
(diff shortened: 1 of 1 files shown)


reana_workflow_controller/rest/utils.py
```diff
@@ -234,14 +234,18 @@ def delete_workflow(workflow, all_runs=False, workspace=False):
...
+                # 1. Stop open interactive sessions
                 int_session = workflow.sessions.first()
                 if int_session:
                     kwrm = KubernetesWorkflowRunManager(workflow)
                     kwrm.stop_interactive_session(int_session.id_)
 
                 if workspace:
+                    # 2. disable retention rules, as the workspace will be deleted
+                    workflow.inactivate_workspace_retention_rules()
+                    # 3. delete the workspace
```
tiborsimko, line 246: Minor comment: wouldn't it be better to change the order of 2 and 3 so that the actual deletion of files will occur first? Thinking about a possible corner case what if something goes wrong with the retention rule inactivation: at least the files would be deleted by the time we get an exception from that function if we switch the order. (Assuming that the most frequent use case is for the researcher to liberate some disk space.)

tiborsimko (approved):
Works well :+1: 

```console
$ grep -C 1 retention reana.yaml
workspace:
  retention_days:
    '*.root': 10
$ reana-client run -w test
$ sleep 60
$ reana-client status -w test | grep -cw finished
1
$ reana-client delete -w test
$ kubectl create job --from cronjob/reana-retention-rules-apply reana-retention-rules-apply2
job.batch/reana-retention-rules-apply2 created
$ kubectl logs reana-retention-rules-apply2-p5d9n
Setting the status of all the rules that will be applied to `pending`
Fetching all the pending rules
No rules to be applied!
```

PR #598, opened 2024-08-08
fix(docker): explicitly specify shell (#598)

### LINKED ISSUE #596: build(docker): fix  JSONArgsRecommended warning
**Current behaviour**

```console
$ reana-dev docker-build -c r-w-controller
...
 => WARN: JSONArgsRecommended: JSON arguments recommended for CMD to prevent unintended behavior related to OS signals (line 89) 
```

See https://docs.docker.com/go/dockerfile/rule/json-args-recommended/

**Expected behaviour**

No warnings should be printed during container image building of REANA components.

mdonadoni:
Fix the `JSONArgsRecommended` warning by explicitly setting the shell to
be used when running the Dockerfile's `CMD` command. Note that it is not
possible to switch to the exec form of `CMD`, as it lacks environment
variable expansion.

Closes #596

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,9 @@
 # Use Ubuntu LTS base image
 FROM docker.io/library/ubuntu:24.04
 
+# Configure shell options
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
 # Use default answers in installation commands
 ENV DEBIAN_FRONTEND=noninteractive
 


Dockerfile
```diff
@@ -85,6 +85,7 @@ EXPOSE 5000
 # Run server
 # exec is used to make sure signals are propagated to uwsgi,
 # while also allowing shell expansion
+SHELL ["/bin/bash", "-c"]
```
tiborsimko, line 88: Could you move the `SHELL` directive into its own paragraph, perhaps somewhere at the top of the `Dockerfile`?  E.g. in the Ntupling Service backend we are using:

```
# Configure shell options
SHELL ["/bin/bash", "-o", "pipefail", "-c"]
```

tiborsimko (approved):

PR #600, opened 2024-08-13
feat(helm): add initial Dask support

Alputer:
This pull request contains the ongoing work of dask integration into REANA.

### DIFF
--- a/reana_workflow_controller/k8s.py
+++ b/reana_workflow_controller/k8s.py
@@ -16,6 +16,7 @@ from reana_commons.k8s.api_client import (
     current_k8s_appsv1_api_client,
     current_k8s_corev1_api_client,
     current_k8s_networking_api_client,
+    current_k8s_custom_objects_api_client,
 )
 from reana_commons.k8s.secrets import UserSecretsStore
 from reana_commons.k8s.volumes import (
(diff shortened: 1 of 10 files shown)


reana_workflow_controller/k8s.py
```diff
@@ -398,3 +399,81 @@ def delete_k8s_ingress_object(ingress_name, namespace):
...
+                "replacement": "/$1",
+            }
+        },
+    }
+
+    ingress = client.V1Ingress(
+        api_version="networking.k8s.io/v1",
+        kind="Ingress",
+        metadata=client.V1ObjectMeta(
+            name=f"dask-dashboard-ingress-{cluster_name}",
```
tiborsimko, line 423: BTW is dashboard exposure working already? I'm getting 404 page not found when trying out locally.

reana_workflow_controller/templates/dask_cluster.yaml
```diff
@@ -0,0 +1,55 @@
...
+              containerPort: 8786
+              protocol: TCP
+            - name: http-dashboard
+              containerPort: 8787
+              protocol: TCP
+          readinessProbe:
+            httpGet:
+              port: http-dashboard
+              path: /health
+            initialDelaySeconds: 5
```
tiborsimko, line 35: It might be good to make these hard-coded options (such as the liveness probe seconds etc) configurable via environment variables. This could come handy for some deployment scenarios. We could also do this later after the initial basic version is merged.

(Alternatively, we could think of customising the whole template, but that would surely go beyond the scope of this initial PR.)

tests/test_dask.py
```diff
@@ -0,0 +1,472 @@
...
+        namespace="default",
+        body={"mock": "cluster_body"},
+    )
+
+
+def test_create_dask_autoscaler(
+    mock_k8s_client, dask_resource_manager, mock_user_secrets
+):
+    # Arrange
+
```
tiborsimko, line 54: Cosmetics: empty line not necessary.

tests/conftest.py
```diff
@@ -152,3 +160,67 @@ def interactive_session_environments(monkeypatch):
         "jupyter",
         {"docker_image_1", "docker_image_2"},
     )
+
+
+#### Fixtures that are needed for `dask.py` but they can also be used later ######
```
tiborsimko, line 165: Comment: do we move the ones that can be of general interest to `pytest-reana`?

reana_workflow_controller/templates/dask_cluster.yaml
tiborsimko: I guess you did not address the comment on hard-coded values. Do we keep this or make the liveness probe etc configurable?

```
            initialDelaySeconds: 5
            periodSeconds: 10
```

(We can keep hard-coded if not terribly important, especially if we use some "safer" values for various deployment scenarios, such as slow cluster vs fast cluster, etc.)

tiborsimko (approved):
Works nicely :+1:

PR #602, opened 2024-09-03
feat(opensearch): fetch logs from opensearch (#602)

jlemesh:
Optionally fetch live workflow and job logs from OpenSearch.

OpenSearch client is configurable via env variables, passed by Helm chart. Defaults are defined in `config.py` file.

Read the Docs CI builds started to fail, added fix from https://about.readthedocs.com/blog/2024/07/addons-by-default/ in a separate commit.

Related to https://github.com/reanahub/reana/pull/827.


jlemesh:
Read the Docs build fails with:
```
      /usr/bin/ld: cannot find core/dot_h.o: No such file or directory
      collect2: error: ld returned 1 exit status
```

tiborsimko (approved):
Works nicely :+1: 

I've generated new `docs/openapi.json` and slightly amended the commit message.

PR #611, opened 2024-11-18
fix(rest): detect session status from the pod state 

Alputer:
Closes reanahub/reana-ui#408

### DIFF
--- a/reana_workflow_controller/k8s.py
+++ b/reana_workflow_controller/k8s.py
@@ -11,6 +11,7 @@ from kubernetes.client.rest import ApiException
 from reana_commons.config import (
     REANA_WORKFLOW_UMASK,
     REANA_RUNTIME_SESSIONS_KUBERNETES_NODE_LABEL,
+    REANA_RUNTIME_KUBERNETES_NAMESPACE,
 )
 from reana_commons.k8s.api_client import (
     current_k8s_appsv1_api_client,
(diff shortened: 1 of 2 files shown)


reana_workflow_controller/k8s.py
```diff
@@ -479,3 +479,16 @@ def delete_dask_dashboard_ingress(cluster_name, workflow_id):
         plural="middlewares",
         name=f"replacepath-{workflow_id}",
     )
+
+
+def check_pod_status_by_prefix(pod_name_prefix, namespace="default"):
```
tiborsimko, line 484: Instead of "default" namespace value, we could use the one we have in `r-commons`, i.e. `REANA_RUNTIME_KUBERNETES_NAMESPACE`.

tiborsimko: Commit message cosmetics: instead of _"get the session status from..."_ we could say "detect session status from..."

tiborsimko: Commit scope cosmetics: since the commit touches component named `k8s.py` or `rest/workflows.py`, we should rather use `fix(rest)` instead of `fix(sessions)`.  The scope is usually named after the component or the subcomponent of the source code that the commit touches; it is not named after the functionality. Although we have not been respecting this conventional commits semantics so for so strictly everywhere :wink:

tiborsimko (approved):

PR #626, opened 2025-01-23
fix(k8s): check readiness of Dask service pods

### LINKED ISSUE #625: Pod Readiness Status Inaccuracy
## Problem Description

Currently, to determine whether our pods are ready to serve requests or still in a pending or non-created state, we rely on the `STATUS` field of the pod object. However, this approach has a subtle inaccuracy. A pod may have a `Running` status even if its readiness probe is unavailable.

We can sometimes see this issue in Dash dashboard badges when Dask service pod is in `Running` state, its `readiness probe` is not available and the Dask badge is visible in the UI. This situation persists for a few seconds.

## Suggested Improvement

To improve reliability, we should rewrite the `check_pod_status_by_prefix` function to depend on the readiness probes of the pods rather than their `STATUS` field. By using readiness probes, we can ensure that the function accurately reflects whether a pod is truly ready to serve requests.

## Current Implementation

    def check_pod_status_by_prefix(
        pod_name_prefix, namespace=REANA_RUNTIME_KUBERNETES_NAMESPACE
    ):
        """Check if there is a Pod in the given namespace whose name starts with the specified prefix. We assume that there exists 0 or 1 pod with a given prefix."""
        try:
            pods = current_k8s_corev1_api_client.list_namespaced_pod(namespace=namespace)
    
            for pod in pods.items:
                if pod.metadata.name.startswith(pod_name_prefix):
                    return pod.status.phase
            return "Not Found"
        except ApiException as e:
            return f"Error: {e.reason}"

Alputer:
Closes #625

### DIFF
--- a/reana_workflow_controller/k8s.py
+++ b/reana_workflow_controller/k8s.py
@@ -1,5 +1,5 @@
 # This file is part of REANA.
-# Copyright (C) 2019, 2020, 2021, 2022, 2024 CERN.
+# Copyright (C) 2019, 2020, 2021, 2022, 2024, 2025 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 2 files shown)


tiborsimko (approved):
Works nicely :+1:

PR #629, opened 2025-01-30
fix(dask): check Traefik before creating dashboard ingress

Alputer:
Before creating the Dask dashboard ingress, the code now checks if either TRAEFIK_ENABLED or TRAEFIK_EXTERNAL is set. This ensures that ingress resources are only created and deleted when Traefik is available.

Closes reanahub/reana#852


tiborsimko (approved):
Works nicely :+1:

PR #630, opened 2025-01-31
fix(config): set localhost default value for REANA_HOSTNAME

Alputer:
This PR is part of harmonizing the treatment of REANA_HOSTNAME accross all REANA components. You can refer to other PRs below.

reanahub/reana#867
reanahub/reana-server#717

Closes reanahub/reana#865

### DIFF
--- a/reana_workflow_controller/config.py
+++ b/reana_workflow_controller/config.py
@@ -26,6 +26,22 @@ def _env_vars_dict_to_k8s_list(env_vars):
     return [{"name": name, "value": str(value)} for name, value in env_vars.items()]
 
 
+def compose_reana_url(hostname: str, hostport: int) -> str:
+    """
+    Compose a REANA API URL while omitting the default HTTPS port (443).
+
+    Args:
+        hostname (str): The REANA hostname.
+        hostport (int): The REANA host port.
+
+    Returns:
+        str: The full base URL.
+    """
+    if hostport == 443:
+        return f"https://{hostname}"
+    return f"https://{hostname}:{hostport}"
+
+
 SECRET_KEY = os.getenv("REANA_SECRET_KEY", "CHANGE_ME")
 """Secret key used for the application user sessions."""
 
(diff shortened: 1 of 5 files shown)


reana_workflow_controller/config.py
```diff
@@ -277,9 +277,12 @@ def _parse_interactive_sessions_environments(env_var):
 REANA_GITLAB_URL = "https://{}".format(REANA_GITLAB_HOST)
 """GitLab API URL"""
 
-REANA_HOSTNAME = os.getenv("REANA_HOSTNAME", "CHANGE_ME")
+REANA_HOSTNAME = os.getenv("REANA_HOSTNAME", "localhost")
 """REANA URL"""
```
tiborsimko, line 281: **todo:** ‏If we take port out of `reana_hostname`, we should change the docstring from "REANA URL" to "REANA host name".

reana_workflow_controller/config.py
```diff
@@ -277,9 +277,12 @@ def _parse_interactive_sessions_environments(env_var):
...
 
+REANA_HOSTPORT = os.getenv("REANA_HOSTPORT", "30443")
+"""REANA Ingress Port"""
```
tiborsimko, line 284: **todo:** ‏Since "host port" and "ingress" do not have apparent literal connection, let's use a docstring such as 
"REANA host name port number". Here but also in Helm value documentation etc.

reana_workflow_controller/consumer.py
```diff
@@ -196,7 +197,9 @@ def _update_commit_status(workflow, status):
         return
     gitlab_access_token = gitlab_access_token_secret.value_str
 
-    target_url = f"https://{REANA_HOSTNAME}/api/workflows/{workflow.id_}/logs"
+    target_url = (
+        f"https://{REANA_HOSTNAME}:{REANA_HOSTPORT}/api/workflows/{workflow.id_}/logs"
```
tiborsimko, line 201: **suggestion:** ‏If `REANA_HOSTPORT` is 443, we don't want to be appending it to HTTPS URLs, because it is more user friendly to write `https://reana.cern.ch` instead of `https://reana.cern.ch:443`.

You could introduce a new utility function that will compose URL of the host name and host port information, leaving the port aside if it has the default 443 value.  (Or use a function from a Python library.)

reana_workflow_controller/config.py
tiborsimko: **polish:** ‏For the release notes, for this component the most important news is that Dask dashboard URLs are correct, so something like the following may be appropriate?

```
fix(dask): use correct REANA host port for Dask service URLs
```

reana_workflow_controller/config.py
```diff
@@ -277,8 +293,14 @@ def _parse_interactive_sessions_environments(env_var):
...
 
-REANA_HOSTNAME = os.getenv("REANA_HOSTNAME", "CHANGE_ME")
-"""REANA URL"""
+REANA_HOSTNAME = os.getenv("REANA_HOSTNAME", "localhost")
+"""REANA host name."""
+
+REANA_HOSTPORT = os.getenv("REANA_HOSTPORT", "30443")
+"""REANA host name port number."""
+
+BASE_REANA_URL = compose_base_reana_url(REANA_HOSTNAME, REANA_HOSTPORT)
```
tiborsimko, line 302: **suggestion:** ‏We can call the variable `REANA_URL`, as we did the past.  Alternatively, `REANA_BASE_URL`, or `REANA_HOST_URL` to make link with host name and host port.  

Please note starting with `REANA_...` rather than `BASE_...`.

(I guess it should be self-understood that it is the base URL of the service, so even `REANA_URL` should be OK to use. Well, we also have `REANA_SERVER_URL`, so one could say we are detecting `REANA_SERVER_URL` from the host name and host port... But rather use something else since that one is usually set by the users. So we may want to keep the distinction of what we detect based on deployment values vs what user sets.)

tiborsimko (approved):

PR #632, opened 2025-02-10
feat(dask): detect and create kerberos resources for Dask cluster

Alputer:
We detect if kerberos is necessary for Dask cluster by checking top level kerberos statement under resources field in `reana.yaml`. No additional changes were needed as first implementation of adding kerberos credentials works just fine.

Closes reanahub/reana#871

### DIFF
--- a/reana_workflow_controller/dask.py
+++ b/reana_workflow_controller/dask.py
@@ -59,6 +59,7 @@ class DaskResourceManager:
         user_id,
         num_of_workers,
         single_worker_memory,
+        kerberos=False,
     ):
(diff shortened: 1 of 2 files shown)


tiborsimko (commented):
Works nicely :+1: 

**todo:** ‏For the release notes, please use Kerberos with uppercase K, with perhaps a little bit more user-oriented language. Perhaps something like:

```
feat(dask): create Kerberos sidecars for Dask workflows (#632)
```

Explaining in the commit message body "If we detect that the user asked for Kerberos resources, we create...".

PR #634, opened 2025-02-11
feat(dask): detect and create rucio resources for Dask cluster

Alputer:
We detect if rucio is necessary for Dask cluster by checking top level rucio statement under resources field in reana.yaml.

Closes reanahub/reana/#873


tiborsimko (approved):
Re-tested VOMS and Kerberos.  Cannot test Rucio, but we can see about this one once deployed on QA.

I'll add a release commit on top whilst merging.

PR #642, opened 2025-03-26
feat(config): introduce new parameters for uWSGI worker recyling

### LINKED ISSUE #641: uwsgi: optimise memory consumption to prevent ingress node freezes
### Current behaviour

Observed a couple of times that the `reana-workflow-controller`'s memory usage was high, which led to the node memory being exhausted and ultimately the node not being accessible, which led to the whole REANA web site not being down, because the `reana-workflow-controller` usually runs on the ingress node.

For reference, here is example of one such situation.

The `reana-workflow-controller` memory consumption as a function of time:

Image

Note that 6 GB is pretty high, and since other pods run on the same node, the system went out of memory.

Kubernetes tries to terminate and restart the pod:

```console
$ kubectl get pods
reana-workflow-controller-77fc84bbc6-hbt8l                        0/2     Pending       0               39m
reana-workflow-controller-77fc84bbc6-r28t6                        2/2     Terminating   4 (7h34m ago)   3d
```

But this was failing due to node being not usable:

```console
$ kubectl get events
17m         Warning   FailedScheduling       pod/reana-workflow-controller-77fc84bbc6-hbt8l                   0/80 nodes are available: 1 node(s) had untolerated taint {node-role.kubernetes.io/master: }, 1 node(s) had untolerated taint {node.kubernetes.io/unreachable: }, 78 node(s) didn't match Pod's node affinity/selector. preemption: 0/80 nodes are available: 80 Preemption is not helpful for scheduling.
```

Indeed the node was not reachable by Kurbernetes:

```console
$ kubectl top nodes | grep node-0
node-0     <unknown>    <unknown>   <unknown>       <unknown>
```

The node could not have been connected to via `ssh` either.

The OpenStack node logs were showing uWSGI memory troubles, such as:

```text
[1861454.463174] Out of memory: Killed process 1287649 (uwsgi) total-vm:6438032kB, anon-rss:6311452kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:12496kB oom_score_adj:1000
[1861454.465107] Out of memory: Killed process 1287578 (uwsgi) total-vm:125800kB, anon-rss:71852kB, file-rss:0kB, shmem-rss:384kB, UID:0 pgtables:276kB oom_score_adj:1000
[1861454.466542] Out of memory: Killed process 1287649 (uwsgi) total-vm:6438032kB, anon-rss:6311452kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:12496kB oom_score_adj:1000
[1861454.468048] Out of memory: Killed process 1287651 (uwsgi) total-vm:4583060kB, anon-rss:4458648kB, file-rss:488kB, shmem-rss:0kB, UID:0 pgtables:8960kB oom_score_adj:1000
[2030357.630346] Out of memory: Killed process 3908681 (uwsgi) total-vm:7346320kB, anon-rss:6929296kB, file-rss:88kB, shmem-rss:0kB, UID:0 pgtables:13732kB oom_score_adj:1000
[2030357.632292] Out of memory: Killed process 3908552 (uwsgi) total-vm:125840kB, anon-rss:71924kB, file-rss:428kB, shmem-rss:384kB, UID:0 pgtables:276kB oom_score_adj:1000
[2030357.633767] Out of memory: Killed process 3908679 (uwsgi) total-vm:4452328kB, anon-rss:4067944kB, file-rss:216kB, shmem-rss:0kB, UID:0 pgtables:8152kB oom_score_adj:1000
```

A soft-reboot via OpenStack made the node to came back, after which the system came back to normal.

### Desired behaviour

If the memory consumption of `reana-workflow-controller` or some other infrastructure pod becomes high, it should not lead to a situation where the node would totally freeze. Two possible solutions:

1. Ideally, we should make sure that the memory consumption of the `reana-workflow-controller` pod remains moderate.  Let's check the uWSGI configuration and optimise its settings so that the memory usage would remain moderate and that the memory would be often vacated when idling.
2. Alternatively, or in addition, we could separate the `reana-worfklow-controller` pod to run on a dedicated node, so that even if the component would run out of memory, it would not disturb the other infrastructure pods. We already do this for DB or MQ pods by means of node labels  (`reana.io/system=infrastructuredb`, `reana.io/system=infrastructuremq`); we could analogously introduce more labels for "fragile" components, such as `reana.io/system=infrastructurewc`, `reana.io/system=infrastructureui`, etc.

Alputer:
Upgrade `uwsgi` version and add new parameters for worker recycling.

- `idle`: Restarts the worker after it has been idle for the specified number of seconds.
- `max-requests`: Restarts the worker after serving the specified number of requests.
- `max-worker-lifetime`: Restarts the worker after running for the specified number of seconds.
- `reload-on-rss`: Restarts the worker when its resident memory exceeds the specified threshold.

Closes #641

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # This file is part of REANA.
-# Copyright (C) 2017, 2018, 2019, 2020, 2021, 2022, 2023, 2024 CERN.
+# Copyright (C) 2017, 2018, 2019, 2020, 2021, 2022, 2023, 2024, 2025 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -73,13 +73,21 @@ ARG UWSGI_BUFFER_SIZE=8192
 ARG UWSGI_MAX_FD=1048576
 ARG UWSGI_PROCESSES=2
 ARG UWSGI_THREADS=2
+ARG UWSGI_IDLE=60
+ARG UWSGI_MAX_REQUESTS=1999
+ARG UWSGI_MAX_WORKER_LIFETIME=3600
+ARG UWSGI_RELOAD_ON_RSS=1024
 ENV FLASK_APP=reana_workflow_controller/app.py \
     PYTHONPATH=/workdir \
     TERM=xterm \
     UWSGI_BUFFER_SIZE=${UWSGI_BUFFER_SIZE:-8192} \
     UWSGI_MAX_FD=${UWSGI_MAX_FD:-1048576} \
     UWSGI_PROCESSES=${UWSGI_PROCESSES:-2} \
-    UWSGI_THREADS=${UWSGI_THREADS:-2}
+    UWSGI_THREADS=${UWSGI_THREADS:-2} \
+    UWSGI_IDLE=${UWSGI_IDLE:-60} \
+    UWSGI_MAX_REQUESTS=${UWSGI_MAX_REQUESTS:-1999} \
+    UWSGI_MAX_WORKER_LIFETIME=${UWSGI_MAX_WORKER_LIFETIME:-3600} \
+    UWSGI_RELOAD_ON_RSS=${UWSGI_RELOAD_ON_RSS:-1024}
 
 # Expose ports to clients
 EXPOSE 5000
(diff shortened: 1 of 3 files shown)


CameronMcClymont:
Replaced by https://github.com/reanahub/reana-workflow-controller/pull/683 which incorporates these changes.

PR #657, opened 2025-07-09
feat(rest): limit workspace file listing and implement GC

### LINKED ISSUE #644: workspace: ensure smooth performance of file listings for workspaces containing 250k files
## Current behaviour

When a workspace contains many produced output files, the `reana-workflow-controller` pod starts to use too much CPU and too much memory. It could happen that the whole system "freezes". This was observed for a workflow that was producing over 250k files, and when the researcher clicked on the "Workspace" tab of the REANA web interface.

(When this happened, the REANA web interface was slow to serve other users as well. The solution was to delete the workflow controller pod that was "stuck" in calculating the workspace listing response.)

## Expected behaviour

We should improve the file serving behaviour for workspaces containing large number of files so that the workflow controller would not "freeze". This could be achieved in several manners:

1. Simply refuse to serve the full file content if there are more than N files in the workspace, and serve only N files. (N should be configurable by the REANA cluster admins, because some deployments may have worker nodes with a lot of memory, some with smaller memory. For example, N could be set to 10k.)
2. Allow specifying filtering, such as `reana-client ls -w myanalysis --filter name=mydirectory`, both from CLI and from Web UI, and restrict serving N files _after_ the filtering. If we do this, then in theory the researcher could repeat downloading all files by issuing many requests, each with desired filtering.
3. Improve pagination so that only parts of the workspace files are returned, i.e. process at most 10 files using some pre-defined sorting order (e.g. creation timestamps), and then move onto next N, so that all the files do not have to be processed in advance. This could have an advantage in case the workspace is modified afterwards manually, since the pagination could be "lost".

Either of these solutions would prevent "freezing" of the workflow controller component.

## Notes

This issue is related to #641

michaelbuchar:
- Introduce FORCE_GARBAGE_COLLECTION to control garbage collection aggressiveness. 
- Add WORKSPACE_DISPLAY_FILE_LIMIT env var to cap workspace file listings. 
- Update list_directory_files to raise BadRequest when the file limit is exceeded.

Closes https://github.com/reanahub/reana-workflow-controller/issues/644

### DIFF
--- a/reana_workflow_controller/config.py
+++ b/reana_workflow_controller/config.py
@@ -18,6 +18,7 @@ from reana_commons.config import (
 )
 from reana_db.models import JobStatus, RunStatus
 from distutils.util import strtobool
+from typing import List
 
 from reana_workflow_controller.version import __version__
 
@@ -465,3 +466,39 @@ CONTAINER_IMAGE_ALIAS_PREFIXES = ["docker.io/", "docker.io/library/", "library/"
 
 MAX_WORKFLOW_SHARING_MESSAGE_LENGTH = 5000
 """Maximum length of the user-provided message when sharing a workflow."""
+
+
+def _parse_comma_separated_list(value: str) -> List[str]:
+    """Parse comma-separated env var values into a list of strings."""
+    if not value:
+        return []
+    return [x.strip() for x in value.split(",") if x.strip()]
+
+
+WORKSPACE_DISPLAY_FILE_LIMIT = int(os.getenv("WORKSPACE_DISPLAY_FILE_LIMIT", "100000"))
+"""Maximum number of file entries returned by workspace listing endpoints."""
+
+_VALID_GC_COMMANDS = {"ls", "list", "rm", "delete"}
+"""Valid FORCE_GARBAGE_COLLECTION command values."""
+
+_gc_env = os.getenv("FORCE_GARBAGE_COLLECTION", "")
+FORCE_GARBAGE_COLLECTION = _parse_comma_separated_list(_gc_env)
+"""Comma-separated list of commands that trigger a manual `gc.collect()` before operations.
+
+Example:
+  $ export FORCE_GARBAGE_COLLECTION=ls,list,rm,delete
+
+Supported values:
+- ls: trigger `gc.collect()` before listing workspace files
+- list: trigger `gc.collect()` before listing all workflows
+- rm: trigger `gc.collect()` before removing workspace files
+- delete: trigger `gc.collect()` before deleting workflows
+"""
+_invalid_gc = sorted(set(FORCE_GARBAGE_COLLECTION) - _VALID_GC_COMMANDS)
+if _invalid_gc:
+    valid_gc_values = ", ".join(sorted(_VALID_GC_COMMANDS))
+    invalid_gc_values = ", ".join(_invalid_gc)
+    raise ValueError(
+        "Invalid FORCE_GARBAGE_COLLECTION values: "
+        f"{invalid_gc_values}. Valid values: {valid_gc_values}"
+    )
--- a/reana_workflow_controller/rest/utils.py
+++ b/reana_workflow_controller/rest/utils.py
@@ -9,7 +9,7 @@
 """REANA Workflow Controller workflows REST API."""
 
 import difflib
-import fs
+import gc
 import json
 import logging
 import mimetypes
@@ -63,6 +63,8 @@ from reana_workflow_controller.config import (
     PROGRESS_STATUSES,
     REANA_GITLAB_HOST,
     PREVIEWABLE_MIME_TYPE_PREFIXES,
+    FORCE_GARBAGE_COLLECTION,
+    WORKSPACE_DISPLAY_FILE_LIMIT,
 )
 from reana_workflow_controller.consumer import _update_workflow_status
 from reana_workflow_controller.errors import (
@@ -72,6 +74,12 @@ from reana_workflow_controller.errors import (
     REANAWorkflowStatusError,
 )
 from reana_workflow_controller.workflow_run_manager import KubernetesWorkflowRunManager
+
+WORKSPACE_FILE_LISTING_FILTER_HINT = (
+    "Please use more specific filters to narrow the results. "
+    "Available filters: file name, size, or last-modified."
+)
+"""Neutral filter hint shown when workspace listings exceed the display limit."""
 
 
 def start_workflow(workflow, parameters):
@@ -222,6 +230,9 @@ def remove_workflow_jobs_from_cache(workflow):
 
 def delete_workflow(workflow, all_runs=False, workspace=False):
     """Delete workflow."""
+    if "delete" in FORCE_GARBAGE_COLLECTION:
+        gc.collect()
+
     if workflow.status in [
         RunStatus.created,
         RunStatus.finished,
@@ -434,8 +445,16 @@ def list_directory_files(
     workspace_path: str, search: Dict[str, List[str]] = None
 ) -> List[dict]:
     """Return a list of files inside a given workspace."""
+    if "ls" in FORCE_GARBAGE_COLLECTION:
+        gc.collect()
     file_list = []
     for file_name in workspace.walk(workspace_path, include_dirs=False):
+        if len(file_list) >= WORKSPACE_DISPLAY_FILE_LIMIT:
+            raise BadRequest(
+                "Too many files to display "
+                f"(limit={WORKSPACE_DISPLAY_FILE_LIMIT}). "
+                f"{WORKSPACE_FILE_LISTING_FILTER_HINT}"
+            )
         st = workspace.lstat(workspace_path, file_name)
         file_info = {
             "name": file_name,
@@ -465,10 +484,12 @@ def remove_files_recursive_wildcard(workspace_path, path_or_pattern):
     :param workspace_path: Directory to delete files from.
     :param path_or_pattern: Wildcard pattern to use for the removal.
     :return: Dictionary with the results:
-       - dictionary with names of succesfully deleted files and their sizes
+       - dictionary with names of successfully deleted files and their sizes
        - dictionary with names of failed deletions and corresponding
        error messages.
     """
+    if "rm" in FORCE_GARBAGE_COLLECTION:
+        gc.collect()
     deleted = {"deleted": {}, "failed": {}}
     for file_name in workspace.glob_or_walk_directory(
         workspace_path, path_or_pattern, topdown=False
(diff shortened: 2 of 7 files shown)


reana_workflow_controller/config.py
```diff
@@ -407,3 +407,12 @@ def _parse_interactive_sessions_environments(env_var):
...
 MAX_WORKFLOW_SHARING_MESSAGE_LENGTH = 5000
 """Maximum length of the user-provided message when sharing a workflow."""
+
+REANA_GC_ENABLED = int(os.getenv("REANA_GC_ENABLED", "0"))
+"""
+Garbage collection aggressiveness level. Valid values are integers from 0 to 9:
+(O: GC disabled, 9: most aggressive GC)
+"""
+
+REANA_MAX_WORKSPACE_FILES = int(os.getenv("REANA_MAX_WORKSPACE_FILES", "100000"))
```
tiborsimko, line 417: The name `REANA_MAX_WORKSPACE_FILES` could seem to imply that a workspace cannot have more than MAX workspace files, while it actually can. (We are only limiting the endpoint reply.)

What about inventing another name that would be more softer? Perhaps `WORKSPACE_FILES_DISPLAY_LIMIT` or `WORKSPACE_DISPLAY_FILE_LIMIT` or somesuch? We can muse IRL about a nice appropriate name...

reana_workflow_controller/rest/utils.py
```diff
@@ -434,8 +434,16 @@ def list_directory_files(
     workspace_path: str, search: Dict[str, List[str]] = None
 ) -> List[dict]:
     """Return a list of files inside a given workspace."""
+    if int(os.getenv("REANA_GC_ENABLED", "0")) > 0:
```
tiborsimko, line 437: This does not have to do `os.getenv()` again; please import the value from the config file. That will be much faster.

reana_workflow_controller/rest/utils.py
```diff
@@ -434,8 +434,16 @@ def list_directory_files(
...
+        gc.collect()
     file_list = []
+    MAX_FILES_DISPLAY = int(os.getenv("REANA_MAX_WORKSPACE_FILES", "100000"))
```
tiborsimko, line 440: DItto here, please just import the value from the config file. (Also everywhere else in this PR where this repeats.)

reana_workflow_controller/config.py
```diff
@@ -407,3 +407,12 @@ def _parse_interactive_sessions_environments(env_var):
 
 MAX_WORKFLOW_SHARING_MESSAGE_LENGTH = 5000
 """Maximum length of the user-provided message when sharing a workflow."""
+
+REANA_GC_ENABLED = int(os.getenv("REANA_GC_ENABLED", "0"))
```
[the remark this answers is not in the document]
michaelbuchar: Sounds good!

reana_workflow_controller/rest/utils.py
```diff
@@ -434,8 +439,15 @@ def list_directory_files(
...
 ) -> List[dict]:
     """Return a list of files inside a given workspace."""
+    if "ls" in FORCE_GARBAGE_COLLECTION:
+        gc.collect()
     file_list = []
     for file_name in workspace.walk(workspace_path, include_dirs=False):
+        if len(file_list) >= WORKSPACE_DISPLAY_FILE_LIMIT:
+            raise BadRequest(
+                f"Too many files to display ({WORKSPACE_DISPLAY_FILE_LIMIT} reached). "
+                "Please narrow down using the `search` filter."
```
tiborsimko: Note that we don't have "search" filter, so the user may be confused:

```console
$ rcg ls -w root6-roofit-yadage-kubernetes                    
==> ERROR: 400 Bad Request: Too many files to display (3 reached). Please narrow down using the `search` filter.

$ rcg ls -w root6-roofit-yadage-kubernetes --filter search=png 
==> ERROR: filter key 'search' is not valid
Available filters are 'name', 'size', 'last-modified'

$ rcg ls -w root6-roofit-yadage-kubernetes --filter name=png  
NAME                                   SIZE    LAST-MODIFIED      
fitdata/plot.png                       15450   2026-02-02T08:01:04
_yadage/yadage_workflow_instance.png   35206   2026-02-02T08:01:33
```

The trouble is obviously that we call the filter "name" in CLI, and "search" in web UI.

We may want to call it something neutral, like "file name filter" in the error message, and perhaps amend web UI to have "name" explictly there, not only "search"...

reana_workflow_controller/rest/utils.py
```diff
@@ -73,6 +75,12 @@
 )
 from reana_workflow_controller.workflow_run_manager import KubernetesWorkflowRunManager
 
+WORKSPACE_FILE_LISTING_FILTER_HINT = (
+    "Please narrow the results using the available filters: "
+    "file name, size, or last-modified."
```
tiborsimko: Nit: when the user is *already* filtering and still hits the limit, "Please narrow the results using the available filters" reads as if they are not filtering yet. Consider changing the wording slightly to work for both cases, e.g.:

```python
WORKSPACE_FILE_LISTING_FILTER_HINT = (
    "Please use more specific filters to narrow the results. "
    "Available filters: file name, size, or last-modified."
)
```

The phrase "more specific" works whether or not filters are already active.

reana_workflow_controller/rest/utils.py
```diff
@@ -434,8 +445,15 @@ def list_directory_files(
     workspace_path: str, search: Dict[str, List[str]] = None
 ) -> List[dict]:
     """Return a list of files inside a given workspace."""
+    if "ls" in FORCE_GARBAGE_COLLECTION:
+        gc.collect()
     file_list = []
     for file_name in workspace.walk(workspace_path, include_dirs=False):
+        if len(file_list) >= WORKSPACE_DISPLAY_FILE_LIMIT:
```
tiborsimko: The limit currently counts only files that pass the search filter (`len(file_list)`), but every walked file still gets an `lstat` call even if it does not match. For a 250k-file workspace with a filter that matches only a few files, all 250k `lstat` calls still happen — the CPU/memory cost of the walk is not bounded.

Consider counting **all walked files** toward the limit, not just matching ones:

```python
file_list = []
walked_count = 0
for file_name in workspace.walk(workspace_path, include_dirs=False):
    walked_count += 1
    if walked_count > WORKSPACE_DISPLAY_FILE_LIMIT:
        raise BadRequest(
            f"Too many files to display (limit={WORKSPACE_DISPLAY_FILE_LIMIT}). "
            f"{WORKSPACE_FILE_LISTING_FILTER_HINT}"
        )
    st = workspace.lstat(workspace_path, file_name)
    file_info = { ... }
    if search:
        if list_files_filter(file_info, search):
            file_list.append(file_info)
    else:
        file_list.append(file_info)
```

This way, a workspace with >100k files always short-circuits after 100k `lstat` calls regardless of the filter, which is the whole point of the protection. The same change would apply to `list_files_recursive_wildcard`.

reana_workflow_controller/config.py
```diff
@@ -465,3 +466,28 @@ def _parse_interactive_sessions_environments(env_var):
...
+
+def _parse_comma_separated_list(value: str) -> List[str]:
+    """Parse comma-separated env var values into a list of strings."""
+    if not value:
+        return []
+    return [x.strip() for x in value.split(",") if x.strip()]
+
+
+WORKSPACE_DISPLAY_FILE_LIMIT = int(os.getenv("WORKSPACE_DISPLAY_FILE_LIMIT", "100000"))
+"""Maximum number of file entries returned by workspace listing endpoints."""
```
tiborsimko, line 479: A typo like `FORCE_GARBAGE_COLLECTION=lis,delet` would silently do nothing. Consider validating the parsed values at config parse time:

```python
_VALID_GC_COMMANDS = {"ls", "list", "rm", "delete"}

_gc_env = os.getenv("FORCE_GARBAGE_COLLECTION", "")
FORCE_GARBAGE_COLLECTION = _parse_comma_separated_list(_gc_env)

_invalid_gc = set(FORCE_GARBAGE_COLLECTION) - _VALID_GC_COMMANDS
if _invalid_gc:
    raise ValueError(
        f"Invalid FORCE_GARBAGE_COLLECTION values: {_invalid_gc}. "
        f"Valid values: {_VALID_GC_COMMANDS}"
    )
```

This would catch misconfiguration at startup rather than letting it pass silently.

tiborsimko:
Nit: consider rewording the commit message for release notes readability. The current headline says "implement GC" which is a bit grand. Suggestion:

```
feat(rest): limit workspace file listing and add forced GC (#657)

Introduce WORKSPACE_DISPLAY_FILE_LIMIT env var to cap the
number of files returned by workspace listing endpoints,
raising BadRequest when the limit is exceeded. Add
FORCE_GARBAGE_COLLECTION env var to allow forcing manual
gc.collect() before selected workspace operations (ls, list, 
rm, delete).

Closes #644
```

reana_workflow_controller/rest/utils.py
michaelbuchar: Fixed

reana_workflow_controller/rest/utils.py
michaelbuchar: Should be fixed!

reana_workflow_controller/config.py
michaelbuchar: Also fixed

reana_workflow_controller/rest/utils.py
tiborsimko: The `walked_count` change I suggested earlier actually introduces a regression: filters become useless on large workspaces. For example, a workspace with >10 files and `--filter name=pt_1` matching only 4 files will still trigger the limit because `walked_count` counts every file in the workspace, not just matching ones.

The problem is that the `search` filter needs file metadata (`lstat`) to decide whether a file matches, so we cannot skip the stat call. But counting all walked files toward the limit means the error message tells users to "use more specific filters" while filters cannot actually help.

Please revert to the original approach of counting only matched files (`len(file_list)`) for the limit check:

```python
file_list = []
for file_name in workspace.walk(workspace_path, include_dirs=False):
    if len(file_list) >= WORKSPACE_DISPLAY_FILE_LIMIT:
        raise BadRequest(...)
    st = workspace.lstat(workspace_path, file_name)
    file_info = { ... }
    if search:
        if list_files_filter(file_info, search):
            file_list.append(file_info)
    else:
        file_list.append(file_info)
```

This way, without filters a 250k-file workspace still gets capped (the protection works), but with filters users can narrow results below the limit — which is exactly what the error message tells them to do. The same revert applies to `list_files_recursive_wildcard`.

Sorry for the bad suggestion — the walked-count tests should also be removed.

michaelbuchar: Should be fixed, would you mind retesting on your example? It's working on my end!

tiborsimko (approved):

PR #673, opened 2026-03-24
Feature/s3 support

etlstrauss:
## Description
This PR implements the possibility to dynamically mount S3 storage via a sidecar and s3fs via /dev/fuse volume by setting the corresponding user variables mentioned in #671 into the interactive session of a workflow. The implementation should allow to disable this functionality via the helm charts / values.yaml file.

### Fixes / Implements
- Implements #671 (partly / only interactive sessions)
#### other issues
- Also implements variable passing for reanahub/reana-job-controller#504

### Dependencies
- [x] Depends on PR reanahub/reana#933 cause the helm charts have to be changed to allow controlling this feature via values.yaml file.

## Type of Change
- New feature (non-breaking change which adds functionality)

## Checklist:
- [ ] Testing was done by using `run-tests.sh`
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] Tests where done by manually applying changes
- [x] New test was added for automatic testing
- [ ] Implementation is completed / ready to be merge into the master branch


etlstrauss:
Addition:
This PR will also be preparing the mounting of s3-storages in the workflow jobs running on k8 backend. This will be done in connection to the repository https://github.com/reanahub/reana-job-controller. See following PR (comming).

sidecars/datastore/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM alpine:3.20
```
tiborsimko, line 1: **🤖 Ship the S3FS datastore image as a released `reana-datastore-s3fs` component (HIGH)**

This PR set adds a complete production image — `Dockerfile`, `app.py`, `mount.py`, `entrypoint.sh`, `requirements.txt` — under `sidecars/datastore/` inside `reana-workflow-controller`, and packages it into the controller's source distribution via `MANIFEST.in`:

```
include sidecars/datastore/Dockerfile
recursive-include sidecars *.py
recursive-include sidecars *.sh
recursive-include sidecars *.txt
```

A controller repository's job is to *construct Kubernetes objects*, not to carry a production image's implementation in its source tree. REANA already has an established convention for sidecar/utility images: standalone repositories that build exactly one image, published as `docker.io/reanahub/<name>:X.Y.Z` and referenced from the consuming components only as a pinned image string (`VOMSPROXY_CONTAINER_IMAGE`, `KRB5_CONTAINER_IMAGE`, …). Those images get their own dedicated build, lint, image-test, and release lifecycle. The datastore sidecar gets none of that here, and forces the `MANIFEST.in` packaging seen above.

This matters more than packaging hygiene because **the image already has two consumers**: `reana-workflow-controller` (interactive sessions, `k8s.py`) *and* `reana-job-controller` (batch jobs, `kubernetes_job_manager.py`), both of which reference `REANA_DATASTORE_IMAGE`. An artefact shared by two components should not be owned by either of them.

There is also a release-completeness defect. The feature declares a sidecar image but never **produces, publishes, versions, or defaults** one: the Helm default is `interactive_sessions.sidecars.datastore.image: ""` (reana PR #933 `values.yaml`), which flows through to `REANA_DATASTORE_IMAGE = ""`. The feature is opt-in (`enabled: false`), so stock deployments are unaffected — but any operator who enables it has no official image to run. As shipped, REANA provides no supported out-of-the-box deployment path: an operator must discover, build, publish, and version the image themselves. That is why this is HIGH rather than a maintainability nit.

On the name: the platform role is already exposed generically as `REANA_DATASTORE_ENABLED` / `REANA_DATASTORE_IMAGE`, the Helm key `sidecars.datastore`, and the `datastore` container name. Naming the repository **`reana-datastore-s3fs`** aligns with that established vocabulary while making the driver explicit, and leaves a clear extension path — a future `reana-datastore-rclone` can supply a compatible image behind the same `REANA_DATASTORE_IMAGE` knob.

Suggested fix — extract the `Dockerfile`, entrypoint, and S3FS implementation into a standalone `reana-datastore-s3fs` repository following the existing image-component pattern (`run-tests.sh` with `--docker-build`/`--docker-test` + hadolint, README, LICENSE, lint config, `ci.yml`), publish a pinned official image, and remove `sidecars/` plus its `MANIFEST.in` entries from `reana-workflow-controller`. Put the pinned official default in the chart (`interactive_sessions.sidecars.datastore.image: docker.io/reanahub/reana-datastore-s3fs:<version>`) rather than in `reana-commons`: the value already propagates Helm → r-w-c → r-j-c at runtime (`workflow_run_manager` injects `REANA_DATASTORE_IMAGE` into the job-controller container env), so a shared Python constant would be redundant and add a needless `reana-commons` release dependency. Keep the Helm `image` setting as an operator override so sites can build and run their own compatible image. Roll this out in order — **release the new image first, repoint the controllers and chart at its pinned tag, then remove the embedded sidecar source** — so the components never reference an image that does not yet exist. The readiness/lifecycle contract that an alternative image must implement is addressed separately, in the related `reana-job-controller` PR #511 review.

reana_workflow_controller/k8s.py
```diff
@@ -224,9 +314,7 @@ def add_environment_variable(self, name, value):
 
     def add_run_with_root_permissions(self):
         """Run interactive session with root."""
-        security_context = client.V1SecurityContext(
-            run_as_user=0, allow_privilege_escalation=False
-        )
+        security_context = client.V1SecurityContext(run_as_user=0, privileged=True)
```
tiborsimko: **🤖 Interactive sessions are made fully `privileged`, not just root (HIGH)**

`add_run_with_root_permissions` (`k8s.py:315`) changed its security context from `V1SecurityContext(run_as_user=0, allow_privilege_escalation=False)` to `V1SecurityContext(run_as_user=0, privileged=True)` (`k8s.py:317`), and this method is called **unconditionally for every Jupyter interactive session** (`k8s.py:420`), regardless of whether the datastore feature is enabled. A `privileged: true` container is granted all Linux capabilities and direct access to host devices — a serious privilege escalation and a container-escape risk — and it makes interactive sessions incompatible with restricted Pod Security Standards. The privileged FUSE requirement belongs solely on the datastore sidecar (already set in `setup_s3_sidecar`, `k8s.py:290-293`), not on the notebook container, and this change is unrelated to S3.

The sidecar itself is correctly gated — `setup_s3_sidecar()`/`setup_s3_storage()` run only when `REANA_DATASTORE_ENABLED and deployment_builder.datastore_enabled` (`k8s.py:408`), the latter true only when the user has `S3_TO_LOCAL_*` secrets — but the privilege escalation is **not**. Consequently, when the datastore feature is disabled, or enabled but the user has no S3 secrets, no datastore sidecar is created, yet every interactive notebook still runs `privileged: true`. This regresses the master-branch default independently of S3 usage. (The batch path does not regress: the main job container retains `allowPrivilegeEscalation: false`, `kubernetes_job_manager.py:267`.)

Suggested fix — revert `add_run_with_root_permissions` to `allow_privilege_escalation=False` (root without privilege escalation, as before) and keep `privileged: true` confined to the datastore sidecar. Document the cluster-level `/dev/fuse` + mount-propagation prerequisite the sidecar needs instead of escalating the user container. Tests — cover both the datastore-disabled and the datastore-enabled-without-S3-secrets cases: no datastore container, no FUSE hostPath / mount propagation, and the session container retains `allowPrivilegeEscalation: false` without `privileged: true`.

sidecars/datastore/app.py
```diff
@@ -0,0 +1,64 @@
...
+    sys.exit(0)
+
+
+def main_logic():
+    """Runs the initial mounting sequence."""
+    global MOUNTING_COMPLETE
+    print("Initializing S3 mounts...")
+    global aliases
+    aliases = mount.mount()
+    MOUNTING_COMPLETE = True
```
tiborsimko, line 49: **🤖 Sidecar reports ready even when S3 mounts fail; interactive sessions have no readiness gate (HIGH)**

In the datastore sidecar, `createS3Mounts` logs `s3fs` failures but never propagates them — a non-zero return code only prints to stderr (`mount.py:48-55`) and `mount()` returns the alias list regardless (`mount.py:90`). `main_logic` then sets `MOUNTING_COMPLETE = True` unconditionally (`app.py:49`), so the `/health` endpoint (`app.py:15`) reports `ready` even when every mount failed. Compounding this, interactive sessions wire **no** Kubernetes readiness/startup probe against the sidecar, so the Jupyter container starts and presents empty or broken mount paths to the user with no error surfaced. (The batch-job readiness gate is the separate finding on reana-job-controller #511; this is about the sidecar's own untruthful signal and the ungated session path.)

Suggested fix — make readiness reflect actual mount state: have `createS3Mounts`/`mount()` collect per-alias success/failure and report ready only when all required mounts succeed (fail fast otherwise). Gate the workload on that signal — a `startupProbe` against a corrected `/health` (native-sidecar approach), or an equivalent shared-volume ready/failed marker — for both sessions and jobs. A correct shared-volume/native lifecycle here also removes the need for the unauthenticated `/shutdown` HTTP endpoint and the hard-coded control port. Tests — readiness reflecting the real mount outcome (success and failure paths).

sidecars/datastore/mount.py
```diff
@@ -0,0 +1,109 @@
...
+
+def createS3Mounts(aliases, base_dir):
+    for i in range(len(aliases)):
+        try:
+            with open(".passwd-s3fs", mode="w", encoding="utf-8") as f:
+                f.write(f"{aliases[i][4]}:{aliases[i][5]}")
+            os.system("chmod 600 .passwd-s3fs")
+            target_path = os.path.join(base_dir, aliases[i][0])
+            target_path = os.path.join(target_path, aliases[i][1])
+            cmd = f"s3fs {aliases[i][1]} {target_path} -o passwd_file=.passwd-s3fs -o url={aliases[i][2]} -o endpoint={aliases[i][3]} -o use_path_request_style -o allow_other"
```
tiborsimko, line 47: **🤖 S3 mount commands are built by shell-interpolating user secrets (HIGH)**

`mount.py` constructs the `s3fs` invocation and its companions with `os.system` and f-string interpolation of user-controlled values — bucket, host, endpoint and the alias-derived target path (`mount.py:47-48`), plus `chmod`/`rm`/`fusermount` (`mount.py:44, 58, 102`). These values come straight from user secrets (`S3_TO_LOCAL_*`), which the server/client secret code does not validate, so a crafted `HOST`/`BUCKET`/alias containing shell metacharacters reaches `os.system` and is injectable — i.e. arbitrary command execution as root inside a privileged `SYS_ADMIN` sidecar, creating a serious container-escape risk. The credential file `.passwd-s3fs` is also written to the relative CWD (`mount.py:42`) rather than a private path.

Suggested fix — replace every `os.system(...)` with `subprocess.run([...], shell=False)` passing arguments as a list so no shell parsing occurs; write the password file to a per-mount path under a private directory (e.g. via `tempfile`) at `0600`, removed in a `finally`. Validate alias/host/bucket values against a strict pattern before use. Tests — a secret value containing shell metacharacters must not execute.

reana_workflow_controller/config.py
```diff
@@ -465,3 +467,21 @@ def _parse_interactive_sessions_environments(env_var):
 
 MAX_WORKFLOW_SHARING_MESSAGE_LENGTH = 5000
 """Maximum length of the user-provided message when sharing a workflow."""
+
+
+REANA_RUNTIME_JOBS_KUBERNETES_TOLERATIONS = os.getenv(
```
tiborsimko, line 490: **🤖 Unrelated tolerations plumbing is added but never consumed (MEDIUM)**

`REANA_RUNTIME_JOBS_KUBERNETES_TOLERATIONS` is added to config (`config.py:472`) and forwarded into the job-controller container env by `workflow_run_manager`, but **nothing in reana-job-controller reads it** — there is no toleration parsing or application to the job pod spec anywhere in that component. The plumbing is dead, and it is orthogonal to the S3/datastore feature (scope creep that complicates review).

Suggested fix — either complete the feature (parse the value in reana-job-controller and apply `tolerations` to the runtime job pod spec, with tests) and split it into its own PR, or drop the config var and the forwarding from this set.

reana_workflow_controller/k8s.py
```diff
@@ -74,16 +77,42 @@ def __init__(
...
+            env=[],
+            volume_mounts=[],
+            ports=[client.V1ContainerPort(container_port=self.port)],
         )
+        containers = [self._session_container]
+        if REANA_DATASTORE_ENABLED:
+            user_secrets = UserSecretsStore.fetch(self.owner_id)
+            all_env = user_secrets.get_env_secrets_as_k8s_spec()
+            s3_env = [
+                s for s in all_env if s.get("name", "").startswith("S3_TO_LOCAL_")
```
tiborsimko, line 193: **🤖 S3 sidecar construction is duplicated across two components (MEDIUM)**

The `S3_TO_LOCAL_*` filtering, env-splitting, sidecar container/volume spec, privileged security context and the readiness/teardown protocol are implemented twice — as raw dicts in reana-job-controller (`kubernetes_job_manager.py`) and as `kubernetes` client objects here in reana-workflow-controller (`k8s.py`). The two already differ (the job path adds a bash readiness wait the session path lacks), which is how such duplication drifts.

Suggested fix — document the datastore sidecar contract first, then factor out shared code only where a genuine common representation emerges. The two controllers legitimately build different object types (raw dicts vs `kubernetes` client objects), so a single shared builder is not required up front; the priority is one authoritative definition of the contract, not premature abstraction.

sidecars/datastore/mount.py
```diff
@@ -0,0 +1,109 @@
...
+
+    pattern = re.compile(r"^S3_TO_LOCAL_(.*)_ALIAS$")
+
+    print("Scanning environment variables for S3 aliases...")
+
+    aliases = []
+    for key, value in os.environ.items():
+        match = pattern.match(key)
+        if match:
+            aliases.append(value)
```
tiborsimko, line 79: **🤖 The `S3_TO_LOCAL_*` secret convention is undocumented, unvalidated, and self-inconsistent (MEDIUM)**

The feature keys entirely off a string-prefix convention — six secrets per alias (`S3_TO_LOCAL_<ALIAS>_{ALIAS,BUCKET,HOST,REGION,ACCESS_KEY,SECRET_KEY}`) — with no schema, no validation, no documentation, and no changelog entry. There is no *first-class* reana-client support (users can add them via the generic `secrets-add --env`, but nothing guides or validates the S3 shape). A mis-named secret fails silently (and, given the readiness finding above, a session starts anyway).

The implementation is also confusing: `mount()` discovers aliases by matching the env-var **name** `S3_TO_LOCAL_(.*)_ALIAS` (`mount.py:71-79`), but then builds the remaining variable names from the matched secret's **value**, not the captured name group, in `getCredentials` (`S3_TO_LOCAL_{value}_BUCKET`, …). The captured name is ignored: `S3_TO_LOCAL_FOO_ALIAS=BAR` makes the implementation look up `S3_TO_LOCAL_BAR_BUCKET` and the related keys. A working configuration must therefore use the alias **value** consistently for the remaining names, contrary to the apparent schema — redundant and error-prone.

Suggested fix — document the convention as the stable v1 input contract and validate it where secrets are read (all six keys present per alias; clear error otherwise). Derive the alias from the variable **name** (the captured group) rather than its value — or validate that name and value agree. Consider first-class reana-client support and, longer term, a structured storage declaration.

reana_workflow_controller/workflow_run_manager.py
```diff
@@ -814,6 +829,18 @@ def _create_job_spec(
...
+                    "name": "REANA_DATASTORE_IMAGE",
+                    "value": json.dumps(REANA_DATASTORE_IMAGE).replace('"', ""),
+                },
+                {
+                    "name": "REANA_DATASTORE_ENABLED",
+                    "value": json.dumps(REANA_DATASTORE_ENABLED),
+                },
+                {
+                    "name": "REANA_DATASTORE_SECRET",
+                    "value": json.dumps(REANA_DATASTORE_SECRET).replace('"', ""),
```
tiborsimko, line 915: **🤖 Optional datastore pull secret becomes a literal `null` secret reference in batch jobs (HIGH)**

When the datastore is enabled without an `imagePullSecret` — the normal optional-secret case — the chart omits the `REANA_DATASTORE_SECRET` env var, so r-w-c reads `REANA_DATASTORE_SECRET = None` (`config.py`). `workflow_run_manager` then serialises it as `json.dumps(None).replace('"', "")` (`workflow_run_manager.py:842`), yielding the literal string `null`, and passes that to reana-job-controller. r-j-c treats the non-empty `"null"` as a configured secret — `if REANA_DATASTORE_ENABLED and REANA_DATASTORE_SECRET != "":` (`kubernetes_job_manager.py:555`) — and appends an image-pull-secret named `null` (`:556`) to every S3-enabled batch Job. This violates the optional setting's contract and makes those Jobs reference a nonexistent secret (harmless warnings for public images, but broken for private registries that rely on a real pull secret).

Suggested fix — fix r-w-c to omit the variable (or pass an empty string) when unset, dropping the `json.dumps(...).replace('"', "")` pattern in favour of the raw string value. Once that producer contract is fixed, r-j-c's guard can be a simple truthiness check. A truthiness check alone is **not** sufficient while r-w-c still emits `"null"` — `"null"` is a truthy string in Python — so if r-j-c must defensively tolerate current inputs independently, it should also explicitly reject the literal `"null"`. Tests — cover both the configured and absent pull-secret cases.

sidecars/datastore/Dockerfile
tiborsimko: @etlstrauss @arm2arm Please let me know whether this philosophy and the proposed name `reana-datastore-s3fs` is OK with you. If yes, I'll create a new (empty) repository of that name under the `reanahub` organisation, so that you will be able to prepare a pull request against it.

arm2arm: I am fine , so you gonna use similar name for sshfs??

tiborsimko: Yes, `reana-datastore-sshfs` could come later if we aim to ship a  separate SSHFS-based datastore sidecar. We could think about `reana-datastore-rclone`, `reana-datastore-webdav` etc.  (We could perhaps aim at a generic `reana-datastore` sidecar with pluggable backends, but such a refactoring could come later.) Do you foresee to have a need for more sidecars soon?

tiborsimko: @arm2arm @etlstrauss I have prepared an initial component skeleton for `reana-datastore-s3fs`, but one  question about the usage of Python vs Go here.

The sidecar is standalone and mostly supervises an external `s3fs` process: validate config, execute mount commands, report readiness, and handle teardown. Since it will run with elevated FUSE privileges, minimising the attack surface matters more here than for an ordinary unprivileged helper container. A small Go binary would avoid shipping Python, pip, Flask, and their dependency tree in this privileged sidecar image. It also makes shell-free argument execution the default via `exec.Command`, which helps avoid the command-injection class of issues discussed above.

Would you be open to implement this sidecar in Go rather than Python? (Keeping the same sidecar contract.) Now is probably the cheapest moment to decide, before the S3FS logic is moved into the new repository.

arm2arm: I agree with you; it will also reduce the size, which means increasing performance.

tiborsimko: Perfect, I'll turn the scaffold into a Golang-based one and submit directly that.

tiborsimko: The new component has now been created: https://github.com/reanahub/reana-datastore-s3fs

PR #679, opened 2026-04-28
refactor(rest): adapt to non-dynamic relationships

tiborsimko:
After dropping `lazy="dynamic"` in reana-db (relationships now return regular lists rather than query objects), update the callers in this repo accordingly.

In `get_workflows`, replace the chained
`user.workflows.filter(...)` /
`user.workflows_shared_with_me.filter(...)` /
`user.workflows.union_all(user.workflows_shared_with_me)` patterns with explicit `Session.query(Workflow).filter(...)` queries. The two reusable building blocks (`owned_workflows` and `shared_with_me`) are constructed once and reused across the branching logic, including the `union_all` case.

Replace `workflow.users_it_is_shared_with.with_entities(...)` with a plain list comprehension over the now-list-shaped relationship; rename the inner loop variable to avoid shadowing the outer `user`.

Convert `.first()` calls to `[0] if rel else None` and `.all()` calls to direct iteration over the list. The class-level `Workflow.sessions.any()` filter expression and `.append()` mutations stay unchanged — both work on either loading style.

Use an explicit `Session.query(Workflow.id_).filter(...).first() is not None` check for `user_has_workflows` to avoid loading the full workflow list just to test for emptiness.

Adjust the matching test assertions in `tests/test_utils.py` and `tests/test_workflow_run_manager.py` to use truthiness on the now-list-shaped `workflow.sessions`.


reana_workflow_controller/rest/utils.py
```diff
@@ -256,7 +256,7 @@ def delete_workflow(workflow, all_runs=False, workspace=False):
                 )
             for workflow in to_be_deleted:
                 # 1. Stop open interactive sessions
-                int_session = workflow.sessions.first()
+                int_session = workflow.sessions[0] if workflow.sessions else None
```
tiborsimko, line 259: Note that CI tests ar failing since we need a new `reana-db` release first.

michaelbuchar:
test_atomic_creation_of_interactive_session() still asserts not sample_serial_workflow_in_db.sessions.all(). After [reana-db#260](https://github.com/reanahub/reana-db/pull/260), sessions becomes a regular list, so this will raise AttributeError instead of testing the intended empty-session condition. The other assertions in this file were already updated to list semantics, I believe this one needs the same change.

reana_workflow_controller/rest/utils.py
tiborsimko: Bumped `reana-db` and all tests are now passing.

tiborsimko:
> test_atomic_creation_of_interactive_session() still asserts not sample_serial_workflow_in_db.sessions.all(). After reana-db#260, sessions becomes a regular list, so this will raise AttributeError instead of testing the intended empty-session condition. The other assertions in this file were already updated to list semantics, I believe this one needs the same change.

Good catch, thanks! Fixed — `sample_serial_workflow_in_db.sessions.all()` is now `sample_serial_workflow_in_db.sessions`, matching the list semantics used in the other tests in this file.

reana_workflow_controller/rest/utils.py
michaelbuchar: Great!

michaelbuchar (approved):
LGTM!

PR #680, opened 2026-05-05
fix(config): harmonise and align REANA host port handling

michaelbuchar:


reana_workflow_controller/config.py
```diff
@@ -349,7 +349,7 @@ def _parse_interactive_sessions_environments(env_var):
 REANA_HOSTNAME = os.getenv("REANA_HOSTNAME", "localhost")
 """REANA host name."""
 
-REANA_HOSTPORT = os.getenv("REANA_HOSTPORT", "30443")
+REANA_HOSTPORT = int(os.getenv("REANA_HOSTPORT", "30443"))
```
tiborsimko: 🤖 **Consider harmonising `compose_reana_url` with reana-server**

reana-server defines its own `compose_reana_url` (added in reanahub/reana-server#767) with a different shape: `REANA_HOSTPORT` is left as the string returned by `os.getenv`, and the function takes `hostport: str | int` and compares with `if str(hostport) == "443":`.

Both implementations produce identical URIs for valid inputs, but the two repos now diverge in style. Suggest aligning them so both components use the same types and the same implementation — either by amending reana-server to match this PR, or by amending this PR to match reana-server.

michaelbuchar: Fixed

tiborsimko (approved):

PR #683, opened 2026-05-19
feat(docker): externalise uWSGI controller configuration

### LINKED ISSUE #952: feat: finish migration to uwsgi and config maps

CameronMcClymont:
Switch reana-workflow-controller and the job-controller pods it spawns
to INI-based uWSGI startup. This replaces the inline workflow-controller
options and the job-controller Flask development server.

Mount the Helm-managed job-controller configuration at /var/reana/uwsgi,
retain Flask startup for debug mode, and align FLASK_DEBUG handling
between the Dockerfile and Python startup logic.

Ship a default workflow-controller configuration for standalone
containers, and bump the pyuwsgi and uwsgitop minimum versions.

Depends on reanahub/reana#953, which provides both production ConfigMaps
and mounts the workflow-controller one. The job-controller mount is
added here.

Closes reanahub/reana#952

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # This file is part of REANA.
-# Copyright (C) 2017, 2018, 2019, 2020, 2021, 2022, 2023, 2024 CERN.
+# Copyright (C) 2017, 2018, 2019, 2020, 2021, 2022, 2023, 2024, 2025, 2026 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -69,43 +69,20 @@ RUN if test -e modules/reana-commons; then \
 # Check for any broken Python dependencies
 RUN pip check
 
+# Install the default uWSGI configuration. Cluster deployments override it
+# by mounting the Helm-managed ConfigMap at the same path.
+COPY etc/uwsgi.ini /var/reana/uwsgi/uwsgi.ini
+
 # Set useful environment variables
-ARG UWSGI_BUFFER_SIZE=8192
-ARG UWSGI_MAX_FD=1048576
-ARG UWSGI_PROCESSES=2
-ARG UWSGI_THREADS=2
 ENV FLASK_APP=reana_workflow_controller/app.py \
     PYTHONPATH=/workdir \
-    TERM=xterm \
-    UWSGI_BUFFER_SIZE=${UWSGI_BUFFER_SIZE:-8192} \
-    UWSGI_MAX_FD=${UWSGI_MAX_FD:-1048576} \
-    UWSGI_PROCESSES=${UWSGI_PROCESSES:-2} \
-    UWSGI_THREADS=${UWSGI_THREADS:-2}
+    TERM=xterm
 
 # Expose ports to clients
 EXPOSE 5000
 
-# Run server
-# exec is used to make sure signals are propagated to uwsgi,
-# while also allowing shell expansion
-# hadolint ignore=DL3025
-CMD exec uwsgi \
-    --buffer-size ${UWSGI_BUFFER_SIZE} \
-    --die-on-term \
-    --hook-master-start "unix_signal:2 gracefully_kill_them_all" \
-    --hook-master-start "unix_signal:15 gracefully_kill_them_all" \
-    --enable-threads \
-    --http-socket 0.0.0.0:5000 \
-    --master \
-    --max-fd ${UWSGI_MAX_FD} \
-    --module reana_workflow_controller.app:app \
-    --need-app \
-    --processes ${UWSGI_PROCESSES} \
-    --single-interpreter \
-    --stats /tmp/stats.socket \
-    --threads ${UWSGI_THREADS} \
-    --vacuum \
-    --wsgi-disable-file-wrapper
+# Run server (use Flask when FLASK_DEBUG is set, otherwise use uWSGI)
+CMD ["/bin/sh", "-c", "case \"$(printf '%s' \"${FLASK_DEBUG}\" | tr '[:upper:]' '[:lower:]')\" in 1|true) exec flask run -h 0.0.0.0 ;; *) exec uwsgi --ini /var/reana/uwsgi/uwsgi.ini ;; esac"]
 
 # Set image labels
 LABEL org.opencontainers.image.authors="team@reanahub.io"
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,5 @@
 # This file is part of REANA.
-# Copyright (C) 2017, 2018, 2019, 2020, 2021, 2024 CERN.
+# Copyright (C) 2017, 2018, 2019, 2020, 2021, 2023, 2024, 2025, 2026 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -24,6 +24,7 @@ recursive-include docs *.py
 recursive-include docs *.png
 recursive-include docs *.md
 recursive-include docs *.txt
+recursive-include etc *.ini
 recursive-include reana_workflow_controller *.html
 recursive-include reana_workflow_controller *.py
 recursive-include reana_workflow_controller/templates *.template
--- a/etc/uwsgi.ini
+++ b/etc/uwsgi.ini
@@ -1,0 +1,20 @@
+[uwsgi]
+stats = /tmp/stats.socket
+http = 0.0.0.0:5000
+module = reana_workflow_controller.app:app
+master = true
+processes = 2
+threads = 2
+enable-threads = true
+single-interpreter = true
+need-app = true
+vacuum = true
+wsgi-disable-file-wrapper = true
+
+buffer-size = 8192
+
+die-on-term = true
+hook-master-start = unix_signal:2 gracefully_kill_them_all
+hook-master-start = unix_signal:15 gracefully_kill_them_all
+
+max-fd = 1048576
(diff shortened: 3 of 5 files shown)


Dockerfile
```diff
@@ -70,42 +70,16 @@ RUN if test -e modules/reana-commons; then \
...
-    --module reana_workflow_controller.app:app \
-    --need-app \
-    --processes ${UWSGI_PROCESSES} \
-    --single-interpreter \
-    --stats /tmp/stats.socket \
-    --threads ${UWSGI_THREADS} \
-    --vacuum \
-    --wsgi-disable-file-wrapper
+# do not use "/bin/sh -c" here as that breaks signal propagation
+CMD ["uwsgi", "--ini", "/var/reana/uwsgi/uwsgi.ini"]
```
michaelbuchar: ### AI-aided review

The Docker image is no longer self-contained. The default CMD now runs uwsgi --ini /var/reana/uwsgi/uwsgi.ini, but this PR does not add that file to the image. The new image will only start when the matching Helm chart mounts the ConfigMap. That also means setting FLASK_DEBUG=1 on the container alone will not switch to Flask; only the Helm debug command override does that.

CameronMcClymont: Regarding the first point, the Docker image doesn't need to be self-contained because it always runs with the Helm deployment which mounts the uwsgi config map, so this is fine.

The second point is true and I've updated this line to be like we have in rjc where setting `FLASK_DEBUG=1` switches it to Flask.

michaelbuchar (dismissed):

michaelbuchar:
I think this is merge-ready with reana#953, only gap I see is to add the runtime-spec tests.

Dockerfile
```diff
@@ -70,42 +70,15 @@ RUN if test -e modules/reana-commons; then \
...
-    --module reana_workflow_controller.app:app \
-    --need-app \
-    --processes ${UWSGI_PROCESSES} \
-    --single-interpreter \
-    --stats /tmp/stats.socket \
-    --threads ${UWSGI_THREADS} \
-    --vacuum \
-    --wsgi-disable-file-wrapper
+# Run server (use Flask when FLASK_DEBUG is set, otherwise use uWSGI)
+CMD ["/bin/sh", "-c", "if [ -n \"${FLASK_DEBUG}\" ]; then exec flask run -h 0.0.0.0; else exec uwsgi --ini /var/reana/uwsgi/uwsgi.ini; fi"]
```
tiborsimko: **🤖 Dockerfile CMD references `/var/reana/uwsgi/uwsgi.ini` but the image never ships it (LOW)**

```dockerfile
CMD ["/bin/sh", "-c", "if [ -n \"${FLASK_DEBUG}\" ]; then exec flask run -h 0.0.0.0; else exec uwsgi --ini /var/reana/uwsgi/uwsgi.ini; fi"]
```

The new CMD launches uWSGI from a config file at `/var/reana/uwsgi/uwsgi.ini`, but the Dockerfile neither `COPY`s nor generates that file. It only exists at runtime as a Kubernetes `ConfigMap` mounted by the Helm chart (added in the sister PR reanahub/reana#953).

In production, this is fine: that PR also overrides `command:` on the pod spec with `["uwsgi", "--ini", "/var/reana/uwsgi/uwsgi.ini"]` *and* mounts the matching ConfigMap on `/var/reana/uwsgi`. The Dockerfile CMD never runs there.

The remaining impact is on **standalone image use** — `docker run reana-workflow-controller:latest` (local dev, CI smoke tests, anyone exercising the image without the full Helm stack) now fails immediately with "unable to load configuration file `/var/reana/uwsgi/uwsgi.ini`". Before this PR, the inline `--buffer-size`/`--processes`/… flags meant a bare `docker run` of the image did not fail due to a missing uWSGI configuration (it may still have required external services or config to be fully functional, but the uWSGI process itself started). This is a small regression in image self-sufficiency.

Suggested fix — ship a default `uwsgi.ini` inside the image so standalone use keeps working; the ConfigMap mount transparently overrides it in cluster deployments:

```dockerfile
COPY etc/uwsgi.ini /var/reana/uwsgi/uwsgi.ini
```

with a checked-in `etc/uwsgi.ini` carrying sensible defaults for the tunables (`processes`, `threads`, `buffer-size`, `max-fd`, …) previously controlled via the now-removed `UWSGI_*` build args. This is non-blocking — production behaviour with PR #953 is unaffected; the fix only restores `docker run` ergonomics.

tiborsimko: **🤖 `FLASK_DEBUG` truthiness check differs between Dockerfile and Python (NIT)**

The two new conditionals interpret `FLASK_DEBUG` differently:

```dockerfile
# Dockerfile
CMD ["/bin/sh", "-c", "if [ -n \"${FLASK_DEBUG}\" ]; then exec flask run -h 0.0.0.0; else exec uwsgi --ini /var/reana/uwsgi/uwsgi.ini; fi"]
```

```python
# reana_workflow_controller/workflow_run_manager.py:_create_job_controller_startup_cmd
if os.getenv("FLASK_DEBUG", "").lower() in ("1", "true"):
    base_cmd = "exec flask run -h 0.0.0.0;"
else:
    base_cmd = "exec uwsgi --ini /var/reana/uwsgi/uwsgi.ini;"
```

The shell `[ -n "${FLASK_DEBUG}" ]` is true for any non-empty string (so `FLASK_DEBUG=0` or `=false` selects Flask), while the Python check only treats `1` / `true` as debug. The inconsistency only surfaces in standalone `docker run` of the workflow-controller image — in production the Helm chart from reanahub/reana#953 overrides `command:` directly and never executes the Dockerfile CMD. Mostly a code-aesthetics nit, kept here only because the Python form (`in ("1", "true")`) is already the established convention in the same file at line 171.

Suggested fix (optional) — align the shell test with the Python semantics so that the two stay consistent if anyone exercises the Dockerfile CMD path directly. The Python check is case-insensitive (`os.getenv(...).lower() in ("1", "true")`), so the shell side should be too:

```dockerfile
CMD ["/bin/sh", "-c", "case \"$(printf '%s' \"${FLASK_DEBUG}\" | tr '[:upper:]' '[:lower:]')\" in 1|true) exec flask run -h 0.0.0.0 ;; *) exec uwsgi --ini /var/reana/uwsgi/uwsgi.ini ;; esac"]
```

A simpler but case-sensitive variant — only matches the lowercase forms `1` and `true`, dropping `True`/`TRUE`/etc. that the Python side would still accept:

```dockerfile
CMD ["/bin/sh", "-c", "if [ \"${FLASK_DEBUG}\" = \"1\" ] || [ \"${FLASK_DEBUG}\" = \"true\" ]; then exec flask run -h 0.0.0.0; else exec uwsgi --ini /var/reana/uwsgi/uwsgi.ini; fi"]
```

tiborsimko: **🤖 PR body doesn't declare the hard dependency on reanahub/reana#953 (MEDIUM)**

The PR body currently reads only `Closes reanahub/reana#952`. It doesn't mention that the new Dockerfile CMD (`exec uwsgi --ini /var/reana/uwsgi/uwsgi.ini`) requires the `uwsgi-config-reana-workflow-controller` ConfigMap and the matching `/var/reana/uwsgi` volume mount, both introduced by the sister PR reanahub/reana#953.

(Note: the explicit `command: ["uwsgi", "--ini", "/var/reana/uwsgi/uwsgi.ini"]` that #953 also adds to the workflow-controller pod spec is *not* a load-bearing dependency — it duplicates what the new Dockerfile CMD already runs. The hard dependency is purely on the ConfigMap and its mount.)

Concretely, if a deployment picks up the new `reana-workflow-controller` image *before* the matching Helm chart from #953 is in place:

- the chart doesn't define `uwsgi-config-reana-workflow-controller`, and it doesn't mount anything on `/var/reana/uwsgi`;
- the new image's CMD runs `exec uwsgi --ini /var/reana/uwsgi/uwsgi.ini`;
- the pod CrashLoopBackOffs with `unable to load configuration file /var/reana/uwsgi/uwsgi.ini`.

This is a real version-skew deployment risk for anyone using `reana-prev-version` images + `reana-next-version` Helm or vice versa across an upgrade window, and for downstream installations that pin chart and image versions independently. The same dependency applies to reana-job-controller#517's image vs. the `uwsgi-config-reana-job-controller` ConfigMap added in #953.

Suggested fix — extend the PR body to make the coordination explicit, e.g.:

```
Closes reanahub/reana#952.

Depends on reanahub/reana#953, which adds the
`uwsgi-config-reana-workflow-controller` ConfigMap and mounts it at
/var/reana/uwsgi. Without that chart change the new Dockerfile CMD
will fail to start (uwsgi cannot load the .ini file). Release ordering:
ship #953 first (or bundle both into the same Helm/image release) so
no deployment ends up with the new image against the old chart.
```

And on the release-tooling side, ensure the next `reana-workflow-controller` image release is gated on the `reana` Helm chart release that contains #953 (the same applies to reana-job-controller#517 and the `uwsgi-config-reana-job-controller` ConfigMap).

tiborsimko: Amended during merge — default `etc/uwsgi.ini` shipped in the image and `COPY`'d to `/var/reana/uwsgi/uwsgi.ini`; the Helm-mounted ConfigMap transparently overrides it in cluster deployments. Folded into squashed commit `bc8e6fd`.

tiborsimko: Amended during merge — Dockerfile CMD now uses the case-insensitive `case`/`tr` form so the shell side matches Python's `.lower() in ("1", "true")`. Folded into squashed commit `bc8e6fd`.

tiborsimko: Amended during merge — PR body and the squashed commit message `bc8e6fd` now declare the dependency on reanahub/reana#953 (provides both production ConfigMaps and mounts the workflow-controller one; the job-controller mount is added here).

PR #684, opened 2026-05-19
feat(config): propagate minimum Kubernetes job user ID env var

### LINKED ISSUE #951: helm: allow cluster admins to configure the minimum accepted `kubernetes_uid` value
## Current behaviour

User-supplied `kubernetes_uid` workflow hint values are accepted as long as they are `>= 100`. This is hard-coded in REANA ecosystem, e.g. in the `reana-job-controller` component. A REANA deployment may want to strengthen the policy to allow only higher user IDs, similar to OpenShift.

  ## Expected behaviour

The minimum accepted `kubernetes_uid` value should be configurable per REANA deployment via the Helm value files, for example by means of an environment variable such as  `REANA_KUBERNETES_JOBS_MIN_USER_UID`.

tiborsimko:
Read `REANA_KUBERNETES_JOBS_MIN_USER_UID` from the environment and forward it to the workflow batch pod so that reana-job-controller can honour the cluster-configured minimum when handling user-supplied `kubernetes_uid` workflow hints.

Closes reanahub/reana#951

### DIFF
--- a/reana_workflow_controller/config.py
+++ b/reana_workflow_controller/config.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2017, 2018, 2019, 2020, 2021, 2022, 2023, 2024, 2025 CERN.
+# Copyright (C) 2017, 2018, 2019, 2020, 2021, 2022, 2023, 2024, 2025, 2026 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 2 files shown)


reana_workflow_controller/config.py
```diff
@@ -194,6 +194,13 @@ def compose_reana_url(hostname: str, hostport: str | int) -> str:
 https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup.
 """
 
+REANA_KUBERNETES_JOBS_MIN_USER_UID = os.getenv("REANA_KUBERNETES_JOBS_MIN_USER_UID")
```
michaelbuchar, line 197: Could we keep the same default here as in `reana-job-controller`, e.g. `"100"`?

`workflow_run_manager.py` forwards this value unconditionally to the job-controller container. If workflow-controller is run outside the updated Helm chart, or if the env var is otherwise unset, this can propagate `None`/an empty value and override job-controller’s own default of `100`. In the worst case, job-controller may then fail while parsing `REANA_KUBERNETES_JOBS_MIN_USER_UID` with `int(...)`.

I think either defaulting this config value to `"100"` here, or only forwarding the env var when it is set, would make the behavior robust across non-Helm/local deployments too.

tiborsimko: Thanks, addressed.

michaelbuchar (approved):

PR #692, opened 2026-06-26
feat(manager): scope workflow and session user secrets

### LINKED ISSUE #978: Allow restricting which user secrets are exposed to workflow jobs and interactive sessions
After #805, user secrets can be propagated to interactive sessions. Today, secrets are generally exposed broadly to workflow job pods, Dask workers, and interactive sessions. This is convenient, but it also means pods receive secrets they do not actually need.

## Problem

There is currently no least-privilege way for a workflow author or an interactive-session caller to restrict which user secrets are exposed to a given pod/session.

We should keep the current default behaviour for backward compatibility, but provide an explicit opt-in mechanism to reduce the exposed secret set when desired.

## Proposed solution

Introduce an optional `secret_names` allowlist.

Behaviour:

- If `secret_names` is omitted: keep current behaviour and expose all user secrets.
- If `secret_names` is an empty list: expose no user secrets.
- If `secret_names` is provided: expose only the named secrets.
- If a requested secret name does not exist: reject the request with a clear validation error.

Implementation approach:

- Filter by secret name once in `UserSecrets`, then derive env refs and mounted file items from the filtered subset.
- Preserve the current behaviour for Kerberos/VOMS/Rucio-enabled paths for now, i.e. do not scope secrets there yet.

## User-facing scope

Interactive sessions:

- `reana-client open ... --secret-name NAME` (repeatable)
- `reana-client open ... --no-secrets`
- `reana-client-go open ... --secret-name NAME` (repeatable)
- `reana-client-go open ... --no-secrets`

Workflow authoring:

- Serial: step-level `secret_names: ["alpha", "beta"]` in `reana.yaml`
- Dask: `workflow.resources.dask.secret_names: ["alpha", "beta"]` in `reana.yaml`
- Snakemake: rule-level `resources: secret_names="alpha,beta"`
- Yadage: step `environment.resources` entry such as `- secret_names: ["alpha", "beta"]`

## Out of scope for this issue

- CWL support
- Secret scoping for Kerberos/VOMS/Rucio-enabled jobs

These can be addressed in follow-up work once the base scoping model is agreed.

## Acceptance criteria

- Existing workflows and interactive-session flows remain unchanged when `secret_names` is not provided.
- Interactive sessions can request either an explicit secret allowlist or no secrets at all.
- Serial, Snakemake, Yadage, and Dask authors can specify the secret subset for their jobs/workers.
- Empty allowlists result in no secrets being exposed.
- Unknown secret names produce a clear user-facing error.
- Kerberos/VOMS/Rucio-enabled paths preserve current behaviour.

## Affected components

- `reana-commons`
- `reana-job-controller`
- `reana-workflow-controller`
- `reana-client`
- `reana-client-go`
- `reana-workflow-engine-serial`
- `reana-workflow-engine-snakemake`
- `reana-workflow-engine-yadage`
- `reana-server`

michaelbuchar:
Scope the run-batch sidecar from declared workflow and step secret usage, preserve feature-managed credentials where required, validate explicit names early, and let notebook sessions inherit workflow defaults when CLI options are omitted.

Closes reanahub/reana#978


reana_workflow_controller/workflow_run_manager.py
```diff
@@ -650,11 +678,44 @@ def _create_job_spec(
...
+        workflow_declared_features = get_declared_workflow_features(
+            self.workflow.get_specification(),
+            workflow_resources,
+        )
+        explicit_workflow_secret_names = get_explicit_workflow_secret_names(
+            self.workflow.get_specification()
+        )
         user_secrets = UserSecretsStore.fetch(owner_id)
+        if explicit_workflow_secret_names:
+            user_secrets.filter_secrets(explicit_workflow_secret_names)
```
CameronMcClymont: The result here is thrown away and the call exists purely so `filter_secrets` raises `REANASecretDoesNotExist` for unknown step-level names at start time. That behaviour is correct, but the code reads like a bug since it leans on this side-effect. Instead, add an explicit `UserSecrets.validate_secret_names(names)` to commons that raises and returns nothing and call that. Or at the minimum, add a comment saying "called for its validation side effect". But I'd lean towards the former.

reana_workflow_controller/dask.py
```diff
@@ -84,23 +84,31 @@ def __init__(
...
-        self.secret_env_vars = self.secrets_store.get_env_secrets_as_k8s_spec()
-        self.secrets_volume_mount = (
-            self.secrets_store.get_secrets_volume_mount_as_k8s_spec()
-        )
         self.kubernetes_uid = WORKFLOW_RUNTIME_USER_UID
 
         self.kerberos = kerberos
         self.voms_proxy = voms_proxy
         self.rucio = rucio
+        self.secret_names = get_workflow_secret_names(self.workflow_resources)
```
CameronMcClymont: With this PR set, a user can write `workflow.resources.dask.secret_names` and it will pass validation, but it won't actually do anything.

The `dask` block has no `additionalProperties: false`, so the key validates fine. But `cluster_spec` (L88) is only ever read for `image` (L90), and the allowlist here comes from the workflow-level `workflow.resources.secret_names` instead. So the key is accepted, ignored without any warnings, and the workers end up with every user secret.

The related issue https://github.com/reanahub/reana/issues/978 specifies this Dask-specific key:

> Dask: `workflow.resources.dask.secret_names: ["alpha", "beta"]` in `reana.yaml`

The implementation needs updated to match what was agreed in the issue: add `secret_names` to the `dask` schema block and read `self.cluster_spec.get("secret_names")` here, falling back to the workflow-level list via `resolve_secret_names()`. This is the same precedence this PR set already uses for serial, snakemake, yadage and CWL steps.

CameronMcClymont (changes requested):

reana_workflow_controller/workflow_run_manager.py
michaelbuchar: Fixed

reana_workflow_controller/dask.py
michaelbuchar: Fixed

reana_workflow_controller/dask.py
```diff
@@ -258,17 +268,19 @@ def _prepare_cluster(self):
...
-        )
+        if self.exposed_secrets.get_secrets():
+            self.cluster_body["spec"]["worker"]["spec"]["containers"][0]["env"].extend(
+                self.secret_env_vars
+            )
 
-        self.cluster_body["spec"]["worker"]["spec"]["containers"][0][
-            "volumeMounts"
-        ].append(self.secrets_volume_mount)
+        if self.exposed_secrets.has_file_secrets():
```
tiborsimko, line 276: **🤖 PR692-01: Env-only scoped Dask Rucio workers reference an absent secret volume (MEDIUM)**

Dask now appends the user-secret volume only when `self.exposed_secrets.has_file_secrets()` is true. However, `_add_rucio_init_container()` always prepends `self.secrets_volume_mount` to the init container. Rucio's required credentials are valid environment-type secrets, so a Dask workflow using Rucio with `secret_names: []` has no file secret: the worker pod omits the user-secret volume while its Rucio init container still references that volume name. The Dask operator's worker pods then fail to be created, so the cluster never becomes ready and the failure surfaces as a stalled workflow rather than a submission error.

Suggested fix — attach the secrets mount to the Rucio init container only when the exposed set contains file secrets. The same construction appears in both deployment-mode branches of `_add_voms_proxy_init_container()`. Valid VOMS configurations normally include file secrets and do not reach the missing-volume case today, but the identical pattern should be corrected in the same pass. Add a regression test that validates the complete Dask worker pod for an env-only scoped Rucio configuration.

reana_workflow_controller/workflow_run_manager.py
```diff
@@ -681,27 +674,101 @@ def _create_job_spec(
...
             )
+        )
+        return user_secrets.filter_secrets(list(dict.fromkeys(sidecar_secret_names)))
+
+    def _get_workflow_user_secrets(self, owner_id):
+        """Return user secrets scoped for workflow engine and job controller."""
+        workflow_resources = self.get_workflow_resources()
+        workflow_specification = self.workflow.get_specification()
+        workflow_secret_names = get_workflow_secret_names(workflow_resources)
+        workflow_sidecar_secret_names = get_declared_workflow_secret_names(
```
tiborsimko, line 703: **🤖 PR692-03: Malformed secret declarations are reported as HTTP 500 (MEDIUM)**

The amended normalization code correctly raises `REANAValidationError` for malformed `secret_names`, and workflow start now calls that code while constructing the workflow pod. However, `reana_workflow_controller/rest/workflows_status.py` only maps `REANASecretDoesNotExist` and `KeyError` to HTTP 400; `REANAValidationError` falls through to its catch-all HTTP 500 response. For example, a malformed Yadage `environment.resources.secret_names` declaration is therefore presented as an internal server failure rather than an actionable input error. This leaves the earlier validation fix only half effective.

Suggested fix — import and handle `REANAValidationError` alongside the other HTTP 400 input errors, and add endpoint-level coverage proving that malformed engine-specific secret declarations return a clear 400 response.

reana_workflow_controller/dask.py
```diff
@@ -114,23 +114,33 @@ def __init__(
...
-        )
         self.kubernetes_uid = int(WORKFLOW_RUNTIME_USER_UID)
 
         self.kerberos = kerberos
         self.voms_proxy = voms_proxy
         self.rucio = rucio
+        self.secret_names = resolve_secret_names(
+            self.cluster_spec.get("secret_names"), self.workflow_resources
+        )
+        self.exposed_secrets = self.secrets_store.get_scoped_secrets(
```
tiborsimko, line 134: **🤖 PR692-04: Dask feature helpers bypass the scoped secret collection (NIT)**

The Dask manager establishes `self.exposed_secrets` as the effective scoped set and uses it for the worker environment and secret volume, but the Kerberos, VOMS, and Rucio helper paths still read from `self.secrets_store`. Feature-required names are currently added to `self.exposed_secrets`, so the results happen to be equivalent today, but bypassing the scoped object weakens the invariant that every consumer sees only the resolved set and makes future divergence easy.

Suggested fix — pass `self.exposed_secrets` consistently to the feature helpers and cover the feature-specific additions in scoped Dask tests.

reana_workflow_controller/workflow_run_manager.py
tiborsimko: **🤖 Follow-up (2026-08-23): verified at current head**

`UserSecrets.validate_secret_names()` now expresses the validation operation directly, and workflow startup calls it instead of discarding a filtered collection for its exception side effect. This addresses the readability and API concern raised here; I consider this discussion resolved.

reana_workflow_controller/dask.py
tiborsimko: **🤖 Follow-up (2026-08-23): verified at current head**

The Dask schema now declares `secret_names`, the resource manager reads `cluster_spec.get("secret_names")`, and `resolve_secret_names()` applies the workflow-global fallback. Regression coverage exercises both the Dask-local override and fallback behavior. This implements the agreed Dask semantics; I consider this discussion resolved.

reana_workflow_controller/workflow_run_manager.py
```diff
@@ -780,227 +982,198 @@ def _create_job_spec(
...
+        """Return the pod security context for workflow pods."""
+        if not K8S_USE_SECURITY_CONTEXT:
+            return None
+
+        return client.V1PodSecurityContext(
+            fs_group=int(WORKFLOW_RUNTIME_USER_GID),
+            fs_group_change_policy=REANA_RUNTIME_FS_GROUP_CHANGE_POLICY,
+        )
+
+    def _pod_volumes(self, workspace_volume, exposed_user_secrets, kerberos):
```
tiborsimko: **🤖 PR692-05: Kerberos init containers can reference an unattached secrets volume (MEDIUM)**

`get_kerberos_k8s_config()` always gives the Kerberos init container a mount for the user's Secret, but `_pod_volumes()` now adds the matching volume only when `exposed_user_secrets.has_file_secrets()` is true. These decisions can disagree for a plain `workflow.resources.kerberos: true` workflow when the user has `CERN_USER` and `CERN_KEYTAB` environment secrets but no file secret matching the filename stored in `CERN_KEYTAB`. No `secret_names` declaration is needed to reach this path.

The generated init container references the user-secret volume while the pod declares only the Kerberos cache and configuration volumes. Kubernetes rejects the Job because the mount names an absent volume, so workflow startup fails at pod submission. Before the conditional volume attachment, the Secret volume was declared and `kinit` could report the missing keytab; the amended path instead produces an invalid pod specification and a less actionable error.

No previously working Kerberos configuration becomes broken: `kinit` already required the referenced keytab file, so configurations reaching this state already failed. The regression is that the failure moves from an actionable init-container error to rejection of an invalid Kubernetes manifest. There is also an asymmetric validation path: a locally scoped job passes feature-required names through `filter_secrets()` and cleanly raises `REANASecretDoesNotExist` for the missing keytab, while sidecar construction uses `get_existing_required_feature_secret_names()` and silently drops it; the unscoped run-batch path skips the filtering validation entirely.

Suggested fix — validate before pod construction that `CERN_KEYTAB` names an existing file-type secret, and raise `REANASecretDoesNotExist` or a dedicated validation error when it does not. Construct the Secret volume and its mount as one unit so callers cannot attach either half independently. Add a complete-pod regression for a Kerberos workflow whose referenced keytab file is missing.

PR #693, opened 2026-06-29
feat(validation): add sandboxed workflow specification validation

### LINKED ISSUE #8: Move workflow specification building and validation to the server
## Background

Today, validating a workflow specification requires the client to call the corresponding workflow engine (CWL, Serial, Snakemake, Yadage) to build the specification, then validate it locally. The server receives this specification and executes it as-is.

## Problem

This client-side approach has two consequences:

- Client language lock-in. Supporting a new client requires reimplementing every supported workflow engine in that language. This is why reana-client-go currently cannot create or submit workflows — porting all engines to Go is impractical.
- No security boundary. The server trusts a specification generated on the client without re-checking it. Client-side validation is a quality-of-life feature that catches mistakes early, not a guarantee the server can rely on.

## Proposal

Move specification building and validation to the server, so any client only needs to send a workflow snapshot (source files) and let the server do the rest.

Because building a specification means evaluating the workflow file, which can execute arbitrary code (e.g. Snakemake run: blocks, Jinja templating), this must happen inside a sandbox with least-privilege access — otherwise the server itself becomes remotely exploitable by a malicious specification.

## Acceptance Criteria

- [ ] Specification building and validation happen server-side, so a client only needs to submit workflow files, not run the corresponding engine
- [ ] Specification building and validation run in a sandboxed, least-privilege environment isolated from the rest of the server
- [ ] The sandbox has network access to fetch external dependencies referenced by the workflow (e.g. git+https CWL tools)

Sohn123:
Spawn a locked-down job to load and validate Snakemake/CWL/Yadage specifications, returning a structured report through the container exit code and logs.

Closes reanahub/reana-workflow-validator#8


reana_workflow_controller/spec_validation.py
```diff
@@ -0,0 +1,441 @@
...
+    except ApiException as e:
+        log.error("Could not read logs for pod %s: %s", pod_name, e)
+        return ""
+
+
+def _parse_report(logs):
+    """Extract the sentinel-wrapped JSON report from the pod logs."""
+    if not logs:
+        return None
+    start = logs.find(REPORT_START)
```
tiborsimko: **🤖 SNDBX-03: Server trusts the sandbox's self-reported verdict — no in-process re-validation of the serialized spec (HIGH)**

*Component:* reana-server — `reana_server/validation.py:201-240,275-316`;
reana-workflow-controller — `reana_workflow_controller/spec_validation.py:117-123,333-346`;
reana-workflow-validator — `reana_workflow_validator/cli.py:120-146`

The sandbox safely *contains* untrusted code execution — that part works. But the
server then trusts the **verdict** that the sandbox computed in a process that had
just executed that untrusted code. `load_and_validate_spec` raises on
`report["valid"] == false` and otherwise stores `report["reana_specification"]`
as authoritative (`reana_server/validation.py:300-316`); `_call_rwc_validate`
returns only `report` and **drops `exit_code`** (`validation.py:240`).

The report is parsed from the sandbox pod's stdout. Untrusted user code runs
*during loading* (`cli.py:122-128`), before the genuine report is emitted
(`cli.py:141-146`), so it can shape the result two trivial ways:

- Print its own sentinel block to stdout. `_parse_report` takes the **first**
  `REPORT_START`/`REPORT_END` it finds (`spec_validation.py:337-341`,
  `logs.find(...)`), so an early forged block wins; since the server reads only
  `report` and ignores `exit_code`, a forged `{"valid": true,
  "reana_specification": <anything>}` is accepted.
- Monkeypatch `validate_serialized_spec`/`sys.stdout` or register an `atexit`
  handler in-process before the genuine emit.

Net effect: an authenticated user can get a spec that violates cluster policy
(non-vetted image, unsupported backend, oversized Dask request, retention beyond
max) accepted and stored as valid — defeating the stated goal that the server
"never trusts a client-serialized specification" (`workflows.py:759`).

**Severity is conditional on intent.** If the validation policy is meant to be a
*security boundary*, this is HIGH. If it is *advisory UX* and the cluster enforces
vetted-images/backends/quota independently at run time, it drops to LOW. That
decision should be made explicitly; the label here reflects the worst case.

The fix already exists in the codebase but is applied only on the restart path:
`clone_workflow` → `validate_loaded_spec` → `validate_serialized_spec`
(`reana_server/utils.py:820`). The create/launch paths just don't re-validate.

**Depends on SNDBX-01:** the in-process re-validation is only safe once
`validate_serialized_spec` is genuinely pure. Note the policy checks themselves
(images, backends, Dask, retention, schema) are *already* pure — only the CWL
parameter step shells out — so a scoped "policy-only" re-validation is viable even
before a full purity cleanup.

Suggested fix — treat the sandbox purely as a *loader* that returns a candidate
serialized spec, then re-run the (pure) policy checks server-side on that
candidate to make the authoritative valid/invalid decision, rather than trusting
`report["valid"]`. This is the only real fix: a process that ran untrusted code
can monkeypatch the in-sandbox checks, `os._exit(0)`, or emit only a forged
block, so nothing read back from it can be trusted on its own. Preferring the
**last** sentinel block in `_parse_report` (not the first) and surfacing
`exit_code` are worthwhile defense-in-depth, but they do **not** close the gap by
themselves — they only raise the bar for forging.

reana_workflow_controller/spec_validation.py
```diff
@@ -0,0 +1,441 @@
...
+
+def _terminated_container_state(pod):
+    """Return the validator container's terminated state, or ``None``."""
+    for status in pod.status.container_statuses or []:
+        if status.name == CONTAINER_NAME and status.state and status.state.terminated:
+            return status.state.terminated
+    return None
+
+
+def _read_pod_logs(pod_name):
```
tiborsimko, line 409: **🤖 SNDBX-08: Controller reads the entire validator pod log into memory — stdout-flood DoS (MEDIUM)**

*Component:* reana-workflow-controller —
`reana_workflow_controller/spec_validation.py:312-330` (called from `:278,283`)

`_read_pod_logs` pulls the validator container's full log and decodes all of it
into a single string:

```python
response = current_k8s_corev1_api_client.read_namespaced_pod_log(
    name=pod_name, ..., _preload_content=False,
)
return response.data.decode("utf-8")
```

There is no `limit_bytes` / `tail_lines` bound. The validator runs untrusted code
(Snakemake/CWL/Yadage loading), and in Kubernetes a container's stdout **and**
stderr are merged into this one log stream. A spec that prints a large volume to
stdout during loading therefore forces the controller to buffer the whole thing
in memory. The container has a memory limit, but container stdout is written to
the node's log file, not charged to the container's memory cgroup — so the limit
does not bound this. The result is a controller-side memory-exhaustion vector
driven by an authenticated user's spec.

Note this interacts with SNDBX-03: the genuine report sits at the *end* of the
log, so a naive `limit_bytes` from the start would truncate it. The robust
approach is to bound the read and extract the report from the tail.

Suggested fix — cap the log read (e.g. `tail_lines`/`limit_bytes` sized to a
generous multiple of the largest expected report), and extract the sentinel-
wrapped report from the bounded tail; treat "no parseable report within the cap"
as an infrastructure failure (→ 500), consistent with SNDBX-07.

reana_workflow_controller/config.py
```diff
@@ -335,6 +335,73 @@ def _parse_interactive_sessions_environments(env_var):
...
 
+
+REANA_WORKFLOW_VALIDATOR_IMAGE = os.getenv(
+    "REANA_WORKFLOW_VALIDATOR_IMAGE",
+    "docker.io/reanahub/reana-workflow-validator:latest",
+)
+"""Image for the sandboxed workflow specification validator (loads + validates
+untrusted workflow specs in a locked-down Job)."""
+
+SPEC_VALIDATION_TIMEOUT = int(os.getenv("REANA_SPEC_VALIDATION_TIMEOUT", "15"))
```
tiborsimko: **🤖 SNDBX-16: 15s sandbox deadline may be too short for remote references and cold image pulls (LOW)**

*Component:* reana-workflow-controller —
`reana_workflow_controller/config.py:346-349`,
`reana_workflow_controller/spec_validation.py:151,236`

`SPEC_VALIDATION_TIMEOUT` defaults to 15 seconds and is applied as the Job's
`activeDeadlineSeconds` (`config.py:346-349`, `spec_validation.py:236`). Two cases
can exceed it for legitimate work:

- Loading a workflow that resolves **remote references over the public internet**
  — a yadage `toplevel` git clone, or a chain of CWL `$import`s — can easily take
  longer than 15s, producing a spurious "did not finish within 15s" infrastructure
  failure for a valid spec. (`SPEC_VALIDATION_ALLOW_EGRESS` defaults to `true`
  precisely to support these.)
- `activeDeadlineSeconds` counts from pod start including image pull, and the
  validator image uses `image_pull_policy="IfNotPresent"` (`spec_validation.py:151`).
  On a node that has not cached the image, the first-ever validation can spend the
  whole budget pulling and be killed as `DeadlineExceeded`.

Both surface as infrastructure failures rather than validation results, so they
are not silently wrong — but the default is tight enough to bite real workflows.

Suggested fix — raise the default (e.g. 60s), and/or document that operators
running remote-reference workflows should increase `REANA_SPEC_VALIDATION_TIMEOUT`;
consider pre-pulling / `imagePullPolicy: Always` semantics so a cold pull does not
consume the validation budget.

reana_workflow_controller/spec_validation.py
```diff
@@ -0,0 +1,536 @@
...
+    except ApiException as e:
+        if e.status != 404:
+            log.error("Could not delete spec validation job %s: %s", job_name, e)
+
+
+def _build_egress_rules():
+    """Build the validator pod's egress allow-list.
+
+    Cluster-internal destinations (other pods/services, the nodes, and the
+    link-local cloud metadata endpoint) are *never* reachable: when egress is
```
tiborsimko: **🤖 PR693-01: Standard NetworkPolicy cannot provide the claimed node isolation (MEDIUM)**

The sandbox documentation treats successful creation of this policy as a
fail-closed security boundary:

```python
# Cluster-internal destinations (other pods/services, the nodes, and the
# link-local cloud metadata endpoint) are *never* reachable
```

That guarantee is stronger than Kubernetes `NetworkPolicy` can provide. The
Kubernetes documentation explicitly says that traffic to and from the node on
which a pod runs is always allowed regardless of the node or pod IP, and that a
`NetworkPolicy` resource has no effect at all unless the cluster's network
plugin implements enforcement. Policies are additive as well, so another policy
can widen the selected pods' egress. Creating or replacing the API object
successfully therefore proves none of the things that `_ensure_egress_network_policy`
calls “fail-closed”. This matters here because the container deliberately runs
arbitrary user-controlled Python/CWL/Snakemake code; same-node kubelet,
`hostNetwork` daemon, or monitoring endpoints remain reachable depending on the
deployment.

Reference: https://kubernetes.io/docs/concepts/services-networking/network-policies/

One fairness note: the code's own "fail-closed" *ordering* — the validation
run is aborted when the policy object cannot be ensured
(`_ensure_egress_network_policy`) — is correct and should stay. What needs
softening is the security-boundary claim built on top of it.

Suggested fix — soften the "never reachable" and security-boundary use of
"fail-closed" to describe what is actually guaranteed: policy creation is a
fail-closed ordering gate, while traffic enforcement depends on a
NetworkPolicy-capable CNI and does not block resident-node access. For the
resident-node exception, a CNI/host-firewall mechanism capable of filtering
host endpoints, or dedicated hardened validator nodes, would close the
residual; setting public egress to false does not. Operator documentation and
an enforcement-level cluster test are covered by
[PR980-01 on reana#980](https://github.com/reanahub/reana/pull/980).

reana_workflow_controller/rest/workflows_validation.py
```diff
@@ -0,0 +1,88 @@
...
+        500:
+          description: Infrastructure failure running the validation sandbox.
+          schema:
+            type: object
+            properties:
+              message:
+                type: string
+    """
+    try:
+        exit_code, report = validate_spec_in_sandbox(bundle_path, timeout)
```
tiborsimko, line 85: **🤖 PR693-02: Synchronous validation can exhaust every controller worker (MEDIUM)**

`validate_workflow_spec` holds a uWSGI request open while
`validate_spec_in_sandbox` polls a Kubernetes Job for up to the validation
deadline plus its scheduling allowance. The chart gives the workflow controller
only two processes with two threads, so four overlapping validations occupy the
entire REST worker pool. The new server-side rate limit is `1/5 second`, while a
normal validation may last 60 seconds: one user can still start four overlapping
calls within 15 seconds, and limits are per user rather than a cluster-wide
in-flight bound. At that point unrelated controller calls queue behind the
validator polls.

```python
exit_code, report = validate_spec_in_sandbox(bundle_path, timeout)
return jsonify({"exit_code": exit_code, "report": report}), 200
```

The rate/quota changes are useful and address the immediate request-flood part
of SNDBX-09, but they do not bound concurrency or remove the blocking worker
architecture.

Suggested fix — make validation an asynchronous job API (`202` plus a job ID
and a result/status endpoint), or move the wait into a bounded worker queue. As
an interim safeguard, enforce explicit per-user and cluster-wide in-flight
limits before creating a Job and return `429`/`503` when full, so validation
cannot consume every workflow-controller request worker. The cluster-wide bound
must use state shared across uWSGI processes and controller replicas; a
process-local semaphore would cover only part of the request pool and would
reset on restart.

reana_workflow_controller/config.py
tiborsimko: **🤖 Verified fixed** at `488c7cb` — the default deadline is now 60 seconds and documented, with a fail-fast path for `DeadlineExceeded`.

reana_workflow_controller/spec_validation.py
tiborsimko: **🤖 Verified fixed** at `488c7cb` — policy is no longer sent to or decided by the sandbox; the report contains only the candidate specification/error, the controller parses the last sentinel block, and reana-server forms the authoritative verdict in-process.

reana_workflow_controller/spec_validation.py
tiborsimko: **🤖 Verified fixed** at `488c7cb` — validator log reads are bounded by configurable `tail_lines` and `limit_bytes`, and a missing report within that cap is classified as an infrastructure error.

reana_workflow_controller/spec_validation.py
Sohn123: Addressed. The documentation, configuration, API description, and tests now state the CNI, additive-policy, and resident-node limitations of standard NetworkPolicy.

reana_workflow_controller/rest/workflows_validation.py
Sohn123: We accepted that as a known limitation of the current implementation. There will be a follow up ticket that converts this to the proven async pattern using db + rabbitmq.

reana_workflow_controller/spec_validation.py
Sohn123: One limitation that remains is that this is not testable with the local kind cluster because the kind CNI does not enforce this type of network policies -> Needs testing in QA

reana_workflow_controller/rest/workflows_validation.py
Sohn123: This is the ticket to track this issue: reanahub/reana-workflow-validator#9

reana_workflow_controller/spec_validation.py
```diff
@@ -0,0 +1,542 @@
...
+    policy = client.V1NetworkPolicy(
+        api_version="networking.k8s.io/v1",
+        kind="NetworkPolicy",
+        metadata=client.V1ObjectMeta(
+            name=NETWORK_POLICY_NAME,
+            namespace=REANA_RUNTIME_KUBERNETES_NAMESPACE,
+        ),
+        spec=client.V1NetworkPolicySpec(
+            pod_selector=client.V1LabelSelector(
+                match_labels={"app": VALIDATOR_APP_LABEL}
```
tiborsimko: **🤖 PR693-05: Release-agnostic selector mixes validator egress policies (MEDIUM)**

The policy name is scoped by `REANA_COMPONENT_PREFIX`, but both the validator
Jobs and the policy selector use the same release-agnostic label:

```python
VALIDATOR_APP_LABEL = "reana-spec-validator"
NETWORK_POLICY_NAME = f"{REANA_COMPONENT_PREFIX}-spec-validator-egress"
...
labels = {"reana_workflow_mode": "spec-validation", "app": VALIDATOR_APP_LABEL}
...
pod_selector=client.V1LabelSelector(
    match_labels={"app": VALIDATOR_APP_LABEL}
)
```

If two REANA releases share a runtime namespace, each release-specific policy
therefore selects both releases' validator pods. Kubernetes NetworkPolicies are
additive: the effective egress is the union of every policy selecting a pod.
For example, a release configured with public egress enabled widens the egress
of validator pods belonging to another release that has explicitly disabled
it. Different blocked-CIDR configurations can leak across releases similarly.
This undermines the per-release security knob specifically for the arbitrary
user-controlled code that the sandbox executes.

Reference: https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors

Suggested fix — add an instance-scoped label (for example an
`app.kubernetes.io/instance` value derived from `REANA_COMPONENT_PREFIX`) to
both the Job template and the NetworkPolicy selector. Keep the generic app
label only for grouping if useful. Add a test with two component prefixes in
one namespace and verify that each policy selects only its own validator pod.

docs/openapi.json
```diff
@@ -463,6 +463,69 @@
         "summary": "Move files within workspace."
       }
     },
+    "/api/workflows/validate": {
+      "post": {
+        "consumes": [
+          "application/json"
+        ],
+        "description": "Spawns a maximally-locked-down Kubernetes Job (the reana-workflow-validator image) that loads the raw spec bundle (mounted read-only from ``bundle_path`` on the shared volume) and emits the resulting serialized specification. The sandbox applies no cluster policy; reana-server validates the returned specification itself. Returns the sandbox exit code and the loader report.",
```
tiborsimko: **🤖 PR693-04: The PR693-01 API-description softening is not in the committed OpenAPI specification (LOW)**

The PR693-01 rework carefully softened the network-isolation claims in the
`validate_workflow_spec` docstring:

```yaml
summary: Load a workflow specification in a hardened sandbox.
description: >-
  Spawns a defense-in-depth Kubernetes Job (the
  reana-workflow-validator image) ...
  Network isolation depends on the cluster CNI enforcing NetworkPolicy;
  standard NetworkPolicy cannot isolate the pod from its resident node.
```

but the committed `docs/openapi.json` was not regenerated and still carries
the old overclaiming text:

```json
"summary": "Load a workflow specification in a locked-down sandbox."
"description": "Spawns a maximally-locked-down Kubernetes Job ..."
```

The OpenAPI JSON is the artifact API consumers actually read, so the honesty
fix has not reached the published contract. It also means the repository will
fail `./run-tests.sh --docs-openapi` (the spec is generated from the inline
docstrings and diffed against `docs/openapi.json`). The CI workflow does run
that check as the first step of its `python-tests` job, but the current
unreleased-commons import failure prevents the job from reaching the OpenAPI
diff; once the dependency is released, this drift becomes the next failure.

Relatedly, the bundled copy in
`reana-commons/reana_commons/openapi_specifications/reana_workflow_controller.json`
does not contain the `/api/workflows/validate` operation at all. Nothing
consumes it through bravado today (reana-server calls the endpoint with plain
`requests`), so this is contract completeness rather than a runtime bug, but
the project rule is that the two copies stay in sync with the runtime spec.

Suggested fix — regenerate `docs/openapi.json` from the live app (the
`./run-tests.sh --docs-openapi` flow), commit it, and copy the regenerated
spec to `reana-commons/reana_commons/openapi_specifications/reana_workflow_controller.json`
on the reana-commons PR so both checked-in copies match the runtime spec. Run
the existing OpenAPI check after the new commons release is available.

reana_workflow_controller/spec_validation.py
```diff
@@ -0,0 +1,566 @@
...
+        namespace=REANA_RUNTIME_KUBERNETES_NAMESPACE,
+        labels=labels,
+    )
+
+    # The bundle is a sub-path of the shared volume, mounted read-only. All
+    # writable paths are ephemeral emptyDirs, so nothing the sandbox does can
+    # touch persistent storage.
+    shared_volume = get_reana_shared_volume()
+    work_volume = client.V1Volume(
+        name="spec-validation-work", empty_dir=client.V1EmptyDirVolumeSource()
```
tiborsimko: **🤖 PR693-06: Validator Job has no ephemeral-storage bound for its writable volumes (MEDIUM)**

The sandbox Job hardens CPU and memory but places no limit on writable local
storage:

```python
work_volume = client.V1Volume(
    name="spec-validation-work", empty_dir=client.V1EmptyDirVolumeSource()
)
tmp_volume = client.V1Volume(
    name="spec-validation-tmp", empty_dir=client.V1EmptyDirVolumeSource()
)
...
resources=client.V1ResourceRequirements(
    requests={"cpu": "100m", "memory": "256Mi"},
    limits={
        "cpu": SPEC_VALIDATION_CPU_LIMIT,
        "memory": SPEC_VALIDATION_MEMORY_LIMIT,
    },
),
```

Neither `emptyDir` sets `size_limit`, and the container declares no
`ephemeral-storage` request or limit. Since the loader deliberately executes
arbitrary Snakemake/CWL/Yadage code, that code can fill the two disk-backed
`emptyDir` volumes (`/validation/work`, `/tmp`) or the container log stream up
to node disk pressure, at which point the kubelet evicts pods on that node —
including unrelated workloads. (The container's own writable layer is not a
vector: `read_only_root_filesystem=True` is set. An `emptyDir` without
`medium: Memory` is node-disk-backed, and both it and container logs count
toward local ephemeral storage.) Redirecting the loader's Python `sys.stdout` (in
`cli.py`) does not prevent `os.write`, subprocess output, or a deliberately
reopened file descriptor 1, and the controller's 100 MiB log-read cap protects
controller memory, not node log storage.

This holds independently of PR794-06 / PR7-02: even a correctly scoped,
regular-file-only snapshot can be expanded by the executing loader code.

Suggested fix — set `size_limit` on both `V1EmptyDirVolumeSource` and add
`ephemeral-storage` to the container's `requests` and `limits`
(env-configurable, mirroring `SPEC_VALIDATION_CPU_LIMIT` /
`SPEC_VALIDATION_MEMORY_LIMIT`). Treat pod eviction or deadline expiry caused by
storage pressure as a validation-service/infrastructure failure, not an invalid
specification. (Eviction without a terminated container, and `DeadlineExceeded`,
are controller-side infrastructure failures rather than a loader exit code.)

docs/openapi.json
tiborsimko: **🤖 PR693-04 follow-up (2026-07-27): verified fixed**

*Component:* reana-workflow-controller — `docs/openapi.json` and the bundled
`reana-commons/reana_commons/openapi_specifications/reana_workflow_controller.json`

The committed OpenAPI description now matches the softened code-level claims.
The `/api/workflows/validate` operation reads "Spawns a defense-in-depth
Kubernetes Job ... Network isolation depends on the cluster CNI enforcing
NetworkPolicy; standard NetworkPolicy cannot isolate the pod from its resident
node." and the summary is "Load a workflow specification in a hardened sandbox"
(previously "maximally-locked-down" / "locked-down"). The bundled reana-commons
copy gained the `/api/workflows/validate` path, and both the endpoint and its
500 response schema now declare the `error_id` correlation field.

Verified fixed at `def9521`. No residual.

reana_workflow_controller/spec_validation.py
tiborsimko: **🤖 PR693-05 follow-up (2026-07-27): verified fixed in runtime code; live test not updated (see PR980-02)**

*Component:* reana-workflow-controller —
`reana_workflow_controller/spec_validation.py` (`_validator_labels`,
`_build_network_policy`, `_build_job`)

The validator Job and the NetworkPolicy selector now share a release-scoped
label set, adding `app.kubernetes.io/instance` (defaulting to
`REANA_COMPONENT_PREFIX`):

```python
def _validator_labels(component_prefix=None):
    return {
        "reana_workflow_mode": "spec-validation",
        "app": VALIDATOR_APP_LABEL,
        VALIDATOR_INSTANCE_LABEL: component_prefix or REANA_COMPONENT_PREFIX,
    }
```

Both the Job template and the policy selector are built from these labels, so
two releases sharing a runtime namespace no longer cross-select each other's
validator pods. `test_network_policy_selects_only_its_release_validator_pods`
verifies each policy selects only its own prefix's pods.

Verified fixed at `def9521`.

Residual — the *live* enforcement script in reana#980
(`scripts/test-spec-validator-network-policy.sh`) was not updated to match the
new three-label selector: its probe pod still carries only
`app: reana-spec-validator` and is therefore no longer selected by the policy,
which breaks the one live proof of egress enforcement. Tracked as new finding
**PR980-02**.

docs/openapi.json
Sohn123: Confirmed fixed. The runtime OpenAPI document and the bundled reana-commons copy now contain the same qualified isolation language and `/api/workflows/validate` contract.

reana_workflow_controller/spec_validation.py
Sohn123: Confirmed fixed, including the residual in reana#980. Validator Jobs and policies share the three release-scoped labels, and the live probe now carries and asserts the same complete selector.

reana_workflow_controller/spec_validation.py
Sohn123: Fixed. Work and temporary files now share one size-limited scratch `emptyDir`; the container also has configurable ephemeral-storage request and limit values. Eviction/no-report and `DeadlineExceeded` paths are classified as validation-service failures.

reana_workflow_controller/rest/workflows_validation.py
tiborsimko: **🤖 PR693-02 follow-up (2026-07-29): accepted as a tracked post-merge limitation**

The synchronous request-worker exhaustion risk remains present, but it is now
explicitly acknowledged and tracked in
[reana-workflow-validator#9](https://github.com/reanahub/reana-workflow-validator/issues/9).
That issue records the relevant acceptance criteria: validation must stop
blocking Flask workers in both reana-server and reana-workflow-controller, and
must continue to support validation without first creating a workspace.

I am happy to treat this as a non-blocking architectural follow-up rather than
requesting the asynchronous DB/RabbitMQ implementation in the current PR set.
No further asynchronous implementation is requested here before merge; the
MEDIUM sandbox-validation availability risk remains intentionally accepted
until issue #9 is implemented.

Issue #9 does not cover the synchronous registry-existence phase identified in
PR794-10, which runs in reana-server after a specification has loaded. That
separate worker-holding path still needs its own bounded fix or tracking. The
missing slow-operation limit on both workflow-start entry points is likewise
tracked separately as PR794-13.

reana_workflow_controller/spec_validation.py
```diff
@@ -0,0 +1,589 @@
...
+    except ApiException as e:
+        if e.status == 409:
+            try:
+                existing_policy = (
+                    current_k8s_networking_api_client.read_namespaced_network_policy(
+                        name=policy.metadata.name,
+                        namespace=REANA_RUNTIME_KUBERNETES_NAMESPACE,
+                    )
+                )
+                policy.metadata.resource_version = (
```
tiborsimko: **🤖 PR693-07: Concurrent NetworkPolicy reconciliation can reject valid validations (MEDIUM)**

Every validation reconciles the same release-scoped NetworkPolicy before
creating its Job. When the initial `create` returns 409, the implementation
reads the existing object, copies its `resourceVersion`, and performs a
replacement:

```python
existing_policy = (
    current_k8s_networking_api_client.read_namespaced_network_policy(...)
)
policy.metadata.resource_version = existing_policy.metadata.resource_version
current_k8s_networking_api_client.replace_namespaced_network_policy(
    name=policy.metadata.name,
    namespace=REANA_RUNTIME_KUBERNETES_NAMESPACE,
    body=policy,
)
```

This read-and-replace sequence is not conflict-safe. In steady state the policy
already exists, so two concurrent validation requests can both receive the
create conflict and read the same `resourceVersion`. The first replacement
succeeds; the second receives another 409, which is converted to
`SpecValidationError`. A valid request consequently returns 500 before its Job
is created. Deletion between the create conflict and the read similarly turns a
recoverable read 404 into a 500.

Suggested fix — make reconciliation idempotent under concurrent callers. First
compare the existing policy's relevant specification with the deterministic
desired policy and return without replacing when they already match; this
eliminates nearly all steady-state writes. When an update is required, retry the
whole create/read/replace reconciliation on bounded 409/404 races. Preserve
fail-closed behaviour for genuine Kubernetes API errors and test both a
conflict-then-refresh sequence and deletion between conflict and read.

Sohn123: Fixed. Reconciliation now returns without writing when the existing policy already matches. Required updates retry the complete create/read/replace sequence on bounded 404/409 races with a refreshed resource version, while unexpected API failures still fail closed. Both race cases are tested.

tiborsimko (approved):

PR #694, opened 2026-07-02
feat(manager): harden workflow run and session pods

michaelbuchar:
Run workflow-engine, job-controller, Dask and session pods as the shared runtime user instead of relying on UID 0.

The run-batch startup path now passes the runtime identity expected by the job-controller bootstrap and relies on fsGroup coordination for shared workspace access instead of a root-side workspace chown step.

Interactive sessions also move from root to the fixed runtime user. Session images therefore need to tolerate a non-root entrypoint and the restricted container security context.

### DIFF
--- a/reana_workflow_controller/config.py
+++ b/reana_workflow_controller/config.py
@@ -482,8 +482,59 @@ def _parse_comma_separated_list(value: str) -> List[str]:
     return [x.strip() for x in value.split(",") if x.strip()]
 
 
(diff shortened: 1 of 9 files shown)


tiborsimko (approved):
Thanks, I have validated the behaviour on the default hostPath deployment: run-batch and user-job security contexts, workflow-engine/job-controller UID/GID, workspace writes, regular workflow execution, and Jupyter sessions.

I think that CephFS/CSI/NFS fsGroup behaviour can be tested post-merge.

PR #696, opened 2026-07-14
auth-rework

### LINKED ISSUE #977: Improve SSO and use JWT for external authentication
Improve SSO allow only one provider and use JWT for external authentication.

Sohn123:
Closes: reanahub/reana#977

### DIFF
--- a/reana_workflow_controller/k8s.py
+++ b/reana_workflow_controller/k8s.py
@@ -346,7 +346,7 @@ def build_interactive_jupyter_deployment_k8s_objects(
     workspace,
     access_path,
     image,
-    access_token=None,
+    session_secret=None,
     cvmfs_repos=None,
     owner_id=None,
     workflow_id=None,
@@ -368,6 +368,8 @@ def build_interactive_jupyter_deployment_k8s_objects(
         a ``404``.
     :param image: Jupyter Notebook image to use, i.e.
         ``jupyter/tensorflow-notebook`` to enable ``tensorflow``.
+    :param session_secret: Random per-session secret used as the notebook
+        access token (never a user credential).
     :param cvmfs_mounts: List of CVMFS repos to make available.
     :param owner_id: Owner of the interactive session.
     :param workflow_id: UUID of the workflow to which the interactive
@@ -386,9 +388,11 @@ def build_interactive_jupyter_deployment_k8s_objects(
         "--notebook-dir='{workflow_workspace}'".format(workflow_workspace=workspace),
         f'--NotebookApp.terminado_settings={{"shell_command": ["/usr/bin/bash", "-c", "cd \'{workspace}\' && bash"]}}',
     ]
-    if access_token:
+    if session_secret:
         command_args.append(
-            "--NotebookApp.token='{access_token}'".format(access_token=access_token)
+            "--NotebookApp.token='{session_secret}'".format(
+                session_secret=session_secret
+            )
         )
     deployment_builder.add_command_arguments(command_args)
     deployment_builder.add_reana_shared_storage()
--- a/reana_workflow_controller/workflow_run_manager.py
+++ b/reana_workflow_controller/workflow_run_manager.py
@@ -11,6 +11,7 @@ import copy
 import json
 import logging
 import os
+import secrets
 from typing import List, Optional
 
 from flask import current_app
@@ -510,6 +511,9 @@ class KubernetesWorkflowRunManager(WorkflowRunManager):
         try:
             access_path = self._generate_interactive_workflow_path()
             workflow_run_name = self._workflow_run_name_generator("session")
+            # Random per-session secret used as the notebook access token,
+            # so that no user credential ever reaches the runtime layer.
+            session_secret = secrets.token_urlsafe(32)
             kubernetes_objects = build_interactive_k8s_objects[
                 interactive_session_type
             ](
@@ -517,7 +521,7 @@ class KubernetesWorkflowRunManager(WorkflowRunManager):
                 self.workflow.workspace_path,
                 access_path,
                 validated_image,
-                access_token=self.workflow.get_owner_access_token(),
+                session_secret=session_secret,
                 cvmfs_repos=self.retrieve_required_cvmfs_repos(),
                 owner_id=self.workflow.owner_id,
                 workflow_id=self.workflow.id_,
@@ -538,6 +542,7 @@ class KubernetesWorkflowRunManager(WorkflowRunManager):
                 path=access_path,
                 type_=interactive_session_type,
                 owner_id=self.workflow.owner_id,
+                session_secret=session_secret,
             )
             self.workflow.sessions.append(int_session)
             current_db_sessions = Session.object_session(self.workflow)
--- a/requirements.txt
+++ b/requirements.txt
@@ -61,7 +61,7 @@ pytz==2026.2              # via bravado-core
 pyuwsgi==2.0.30.post1     # via reana-workflow-controller (setup.py)
 pyyaml==6.0.3             # via bravado, bravado-core, kubernetes, reana-commons, swagger-spec-validator
 reana-commons[kubernetes]==0.95.0a20  # via reana-db, reana-workflow-controller (setup.py)
-reana-db==0.95.0a10       # via reana-workflow-controller (setup.py)
+reana-db==0.95.0a11       # via reana-workflow-controller (setup.py)
 requests==2.34.2          # via bravado, bravado-core, kubernetes, opensearch-py, reana-workflow-controller (setup.py), requests-oauthlib
 requests-oauthlib==2.0.0  # via kubernetes
 rfc3987==1.3.8            # via jsonschema
--- a/requirements.txt
+++ b/requirements.txt
@@ -61,7 +61,7 @@ pytz==2026.2              # via bravado-core
 pyuwsgi==2.0.30.post1     # via reana-workflow-controller (setup.py)
 pyyaml==6.0.3             # via bravado, bravado-core, kubernetes, reana-commons, swagger-spec-validator
 reana-commons[kubernetes]==0.95.0a20  # via reana-db, reana-workflow-controller (setup.py)
-reana-db==0.95.0a11       # via reana-workflow-controller (setup.py)
+reana-db==0.95.0a10       # via reana-workflow-controller (setup.py)
 requests==2.34.2          # via bravado, bravado-core, kubernetes, opensearch-py, reana-workflow-controller (setup.py), requests-oauthlib
 requests-oauthlib==2.0.0  # via kubernetes
 rfc3987==1.3.8            # via jsonschema
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,7 @@ extras_require = {
         "apispec[yaml]>=3.0",
         "apispec-webframeworks",
         "reana-commons[kubernetes,tests]>=0.95.0a20,<0.96.0",
-        "reana-db[tests]>=0.95.0a10,<0.96.0",
+        "reana-db[tests]>=0.95.0a11,<0.96.0",
     ],
 }
 
@@ -55,7 +55,7 @@ install_requires = [
     "opensearch-py>=2.7.0,<2.8.0",
     "packaging>=18.0",
     "reana-commons[kubernetes]>=0.95.0a20,<0.96.0",
-    "reana-db>=0.95.0a10,<0.96.0",
+    "reana-db>=0.95.0a11,<0.96.0",
     "requests>=2.25.0",
     "sqlalchemy-utils>=0.31.0",
     "uwsgi-tools>=1.1.1",
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,7 @@ extras_require = {
         "apispec[yaml]>=3.0",
         "apispec-webframeworks",
         "reana-commons[kubernetes,tests]>=0.95.0a20,<0.96.0",
-        "reana-db[tests]>=0.95.0a11,<0.96.0",
+        "reana-db[tests]>=0.95.0a10,<0.96.0",
     ],
 }
 
@@ -55,7 +55,7 @@ install_requires = [
     "opensearch-py>=2.7.0,<2.8.0",
     "packaging>=18.0",
     "reana-commons[kubernetes]>=0.95.0a20,<0.96.0",
-    "reana-db>=0.95.0a11,<0.96.0",
+    "reana-db>=0.95.0a10,<0.96.0",
     "requests>=2.25.0",
     "sqlalchemy-utils>=0.31.0",
     "uwsgi-tools>=1.1.1",
--- a/tests/test_workflow_run_manager.py
+++ b/tests/test_workflow_run_manager.py
@@ -423,3 +423,25 @@ def test_create_job_spec_drops_colliding_job_controller_env_vars(
     assert not any(
         "workflow_runtime_user_uid" in message for message in warning_messages
     )
+
+
+def test_interactive_session_uses_per_session_secret(sample_serial_workflow_in_db):
+    """Per-session random secret is stored and used as the notebook token."""
+    with patch.multiple(
+        "reana_workflow_controller.k8s",
+        current_k8s_corev1_api_client=DEFAULT,
+        current_k8s_networking_api_client=DEFAULT,
+        current_k8s_appsv1_api_client=DEFAULT,
+    ) as mocks:
+        kwrm = KubernetesWorkflowRunManager(sample_serial_workflow_in_db)
+        kwrm.start_interactive_session(
+            InteractiveSessionType(0).name, expose_secrets=False
+        )
+        int_session = sample_serial_workflow_in_db.sessions[0]
+        secret = int_session.session_secret
+        assert secret
+        assert len(secret) >= 32
+        deployment_call = mocks[
+            "current_k8s_appsv1_api_client"
+        ].create_namespaced_deployment.call_args
+        assert f"--NotebookApp.token='{secret}'" in repr(deployment_call)


reana_workflow_controller/k8s.py
```diff
@@ -386,9 +388,11 @@ def build_interactive_jupyter_deployment_k8s_objects(
         "--notebook-dir='{workflow_workspace}'".format(workflow_workspace=workspace),
         f'--NotebookApp.terminado_settings={{"shell_command": ["/usr/bin/bash", "-c", "cd \'{workspace}\' && bash"]}}',
     ]
-    if access_token:
+    if session_secret:
         command_args.append(
-            "--NotebookApp.token='{access_token}'".format(access_token=access_token)
+            "--NotebookApp.token='{session_secret}'".format(
```
tiborsimko: **🤖 PR696-01: Notebook token is exposed in the pod specification (LOW)**

The new per-session secret is injected as a container command argument (`reana_workflow_controller/k8s.py:391-396`):

```python
command_args.append(
    "--NotebookApp.token='{session_secret}'".format(
        session_secret=session_secret
    )
)
```

Command arguments are stored verbatim in the Deployment and Pod objects. Anyone allowed to read those objects in the runtime namespace can therefore recover a live notebook token without permission to read Kubernetes Secrets. Replacing the former account credential with a session-scoped token greatly reduces the blast radius, but the remaining disclosure crosses an avoidable RBAC boundary.

Suggested fix — create a per-session Kubernetes Secret and expose only a `secretKeyRef` or mounted secret file to the notebook container, using the token environment/file mechanism supported by the permitted notebook image. Delete the Secret with the session's other Kubernetes objects.

tiborsimko: This PR has not changed since the review (head is still `b1ee8a8`), so this finding stands as-is. The per-session secret design itself is sound (high entropy, encrypted at rest, deleted with the session); the remaining gap is the pod-spec exposure described above.

reana_workflow_controller/k8s.py
```diff
@@ -346,7 +400,7 @@ def build_interactive_jupyter_deployment_k8s_objects(
     workspace,
     access_path,
     image,
-    access_token=None,
+    session_secret=None,
```
tiborsimko: **🤖 PR696-03: Notebook credential delivery is optional and its image contract is unenforced (LOW)**

`build_interactive_jupyter_deployment_k8s_objects()` declares
`session_secret=None` and adds the managed Secret/`NOTEBOOK_ARGS` only under
`if session_secret`. The manager currently generates a high-entropy value, but
the builder does not enforce that invariant. A future or direct caller can
silently deploy a session whose database/UI secret opens nothing while Jupyter
generates a different token and exposes it through its startup output.

Delivery now also depends on the selected image's `start-notebook.sh`
honouring the Jupyter Docker Stacks `NOTEBOOK_ARGS` convention. Custom images
are supported by configuration, but this new credential-delivery requirement
is not part of their documented/validated contract. An image that starts
Jupyter but ignores `NOTEBOOK_ARGS` fails closed, yet leaves an opaque,
undiagnosable token mismatch.

Suggested fix — make `session_secret` a required non-empty argument, remove
the conditional fallback, and document the `NOTEBOOK_ARGS` compatibility
requirement for every recommended/allowed custom Jupyter image. Add a builder
test that omission/empty input is rejected and an administrator-facing startup
error or validation for an incompatible image where practical.

reana_workflow_controller/k8s.py
```diff
@@ -386,11 +442,9 @@ def build_interactive_jupyter_deployment_k8s_objects(
         "--notebook-dir='{workflow_workspace}'".format(workflow_workspace=workspace),
         f'--NotebookApp.terminado_settings={{"shell_command": ["/usr/bin/bash", "-c", "cd \'{workspace}\' && bash"]}}',
     ]
-    if access_token:
-        command_args.append(
-            "--NotebookApp.token='{access_token}'".format(access_token=access_token)
-        )
     deployment_builder.add_command_arguments(command_args)
+    if session_secret:
```
tiborsimko: **🤖 PR696-04: Moving the notebook token out of pod arguments breaks automatic session cleanup (MEDIUM)**

The amended controller deliberately removes `--NotebookApp.token` from the
container arguments and injects it through a Secret-backed `NOTEBOOK_ARGS`;
the new test asserts that no container argument contains `NotebookApp.token`.
reana-server's `interactive-session-cleanup` command still obtains the token
with `next(filter(lambda a: "token" in a, container_args)).split("'")[1]`.

For every session created by this PR, that expression raises `StopIteration`.
The command's broad exception handler logs the parse failure and continues, so
the inactivity CronJob silently skips every new notebook: pods keep consuming
resources and their session credentials remain valid until a user or workflow
lifecycle action closes them manually.

Suggested fix — ship the integration fix with this controller change. Resolve
the workflow from the pod's existing workflow-UUID label and read the
encrypted `InteractiveSession.session_secret`, or read the named per-session
Kubernetes Secret, instead of scraping container arguments. Add a
cross-component regression test using the new token-free Deployment. Also pin
the lifecycle invariant that the Ingress remains the first owner root and
that Secret/Service/Deployment owner references cascade from it; an
intermediate amendment already broke that ordering before the current head
restored it. Cross-reference PR789-25 for the separate revocation policy.

reana_workflow_controller/k8s.py
```diff
@@ -386,11 +442,9 @@ def build_interactive_jupyter_deployment_k8s_objects(
...
+        deployment_builder.add_session_secret(session_secret)
```
tiborsimko: **🤖 PR696-05: A user `NOTEBOOK_ARGS` secret can replace the managed notebook token (LOW)**

The builder appends the managed Secret-backed `NOTEBOOK_ARGS` entry before
calling `add_user_secrets()`. User environment secrets are appended afterwards
without filtering controller-owned names. Kubernetes resolves duplicate
container environment names by the later value, so an environment secret
named `NOTEBOOK_ARGS` replaces the managed token argument.

`NOTEBOOK_ARGS` is itself a documented Jupyter Docker Stacks customization
knob, making accidental collision plausible. The usual result is that the
server/UI-held session secret no longer works; the worst case is a user value
such as `--NotebookApp.token=''`, which starts an unauthenticated notebook on
the public workflow ingress. The batch/job-controller path already filters
operator-supplied collisions with controller-managed environment variables.

Suggested fix — treat `NOTEBOOK_ARGS` as controller-owned for interactive
sessions and reject or ignore a colliding user secret with an explicit
warning, mirroring `_filter_job_controller_env_var_collisions`. Add a test with
a user environment secret named `NOTEBOOK_ARGS` and assert the managed
`secretKeyRef` remains authoritative.

reana_workflow_controller/k8s.py
tiborsimko: **🤖 PR696-01 follow-up (2026-07-28): verified fixed**

*Component:* reana-workflow-controller — notebook session credential handling

The user's bearer token is no longer placed in Deployment/container arguments.
The controller generates a random per-session token, stores it encrypted in
the session row, creates a Kubernetes Secret containing `NOTEBOOK_ARGS`, and
references that Secret through `secretKeyRef`. The Secret participates in the
session owner chain and lifecycle cleanup.

Verified fixed at `c43b092`. The still-current private draft PR696-03 concerns
making this secret mandatory at the builder boundary; it is separate from
exposure of the user credential.

reana_workflow_controller/k8s.py
tiborsimko: **🤖 PR696-03 follow-up (2026-08-11): verified fixed at 595edf4**

Interactive deployment construction now rejects an empty session secret rather
than silently producing an unauthenticated notebook. The builder documents that
the secret is the notebook access token, and the implementation records the
Jupyter Docker Stacks `NOTEBOOK_ARGS` contract used to consume it from a
Kubernetes Secret. Tests cover the required invariant and secret-backed
environment wiring, closing the silent fail-open path.

reana_workflow_controller/k8s.py
tiborsimko: **🤖 PR696-04 follow-up (2026-08-11): verified fixed across reana-server 6320b65 and reana-workflow-controller 595edf4**

The cleanup rewrite in reana-server now uses the pod's workflow and user labels
to look up the persisted `InteractiveSession`, decrypt its dedicated session
secret, and call the notebook shutdown endpoint without a user access token.
The workflow-controller head supplies and persists that per-session secret, and
the coordinated server regression test verifies token-free controller
arguments. The cleanup path therefore remains authenticated after the general
REANA token is removed from the runtime layer.

reana_workflow_controller/k8s.py
tiborsimko: **🤖 PR696-05 follow-up (2026-08-11): verified fixed at 595edf4**

`NOTEBOOK_ARGS` is now a controller-reserved interactive-session environment
name and user-provided secrets with that name are filtered before they are
appended. Regression coverage verifies that the single surviving entry is the
managed Secret reference carrying the notebook token. A user secret can no
longer override the session authentication argument through Kubernetes's
last-duplicate-wins behaviour.

reana_workflow_controller/k8s.py
```diff
@@ -36,6 +38,46 @@
...
+    permanently and silently excluded from every interactive session with no
+    signal to notice, let alone fix, the collision.
+    """
+    colliding_names = sorted(
+        env_var["name"]
+        for env_var in env_vars
+        if env_var["name"] in RESERVED_SESSION_ENV_VARS
+    )
+    if colliding_names:
+        raise REANAInteractiveSessionError(
```
tiborsimko: **🤖 PR696-06: Reserved-name user secrets hard-fail session start as HTTP 500 (MEDIUM)**

Changing the collision handling from drop-and-warn to an explicit rejection is
useful, but `REANAInteractiveSessionError` reaches
`open_interactive_session()`'s bare `except Exception` and is returned as HTTP
500. A user-fixable input conflict is therefore reported as an internal server
failure.

The behaviour is also an abrupt compatibility change: any existing user with
a secret called `NOTEBOOK_ARGS`, `NB_GID`, `NB_UMASK` or `REANA_WORKSPACE` can
no longer start any interactive session, with no advance signal or upgrade
guidance. These names can be legitimate for batch workflows, so rejecting them
globally when secrets are uploaded would be unnecessarily broad.

Suggested fix — catch `REANAInteractiveSessionError` explicitly at the
interactive-session endpoint and return 400 with the actionable collision
message. Document the upgrade impact and add coverage for an existing
conflicting secret, while continuing to allow the same secret names in batch
workflows.

tiborsimko: **🤖 PR696-06 follow-up (2026-08-21): partially addressed at b2e0884**

The reserved-name exception is now preserved through session creation and
handled by the REST endpoint as a user-correctable 400 with the actionable
secret-name message rather than a generic 500. Existing conflicting secrets
are covered by tests, and the restriction remains specific to interactive-
session creation rather than batch workflows. This resolves the functional
500 reported by the finding.

The upgrade impact remains undocumented. The chart README states that these
environment-variable names are reserved, but it does not tell operators that
an existing user who already has `NOTEBOOK_ARGS`, `NB_GID`, `NB_UMASK`, or
`REANA_WORKSPACE` in their secrets store will be unable to open an interactive
session until that secret is renamed or removed.

Suggested fix — add an upgrade note naming the four reserved secrets and the
required remediation for existing users. Keep the current 400 response and
batch-workflow compatibility unchanged.

Sohn123: Fixed, including the follow-up's remaining ask. `rest/workflows_session.py`'s `open_interactive_session` now catches `ReservedEnvironmentVariableError` and returns `400` with the collision message, instead of falling through to an unclassified `500`. The follow-up's request for upgrade-facing documentation is also done: the reana (Helm) repo's `README.md` names the four reserved secret names (`NOTEBOOK_ARGS`, `NB_GID`, `NB_UMASK`, `REANA_WORKSPACE`) and tells operators to rename any colliding user secrets before upgrading.

reana_workflow_controller/workflow_run_manager.py
```diff
@@ -767,9 +798,19 @@ def _create_job_spec(
...
-                    http_get=client.V1HTTPGetAction(
-                        port=JOB_CONTROLLER_CONTAINER_PORT,
-                        path=JOB_CONTROLLER_SHUTDOWN_ENDPOINT,
+                    _exec=client.V1ExecAction(
+                        command=[
+                            "python3",
+                            "-c",
+                            (
+                                "import urllib.request; "
+                                "urllib.request.urlopen("
```
tiborsimko: **🤖 PR696-07: The loopback preStop request honours proxy environment variables (LOW)**

The new exec-based preStop hook correctly moves the shutdown request inside the
job-controller container, but it uses the process-global urllib opener:

```python
urllib.request.urlopen(
    "http://127.0.0.1:5000/shutdown", timeout=10
).read()
```

`urlopen()` installs a `ProxyHandler` populated from `HTTP_PROXY`/`http_proxy`
and bypasses it only when the environment's `NO_PROXY` covers `127.0.0.1`.
Sites can pass arbitrary job-controller environment variables through
`components.reana_job_controller.environment`: the chart serialises that value
as `REANA_JOB_CONTROLLER_ENV_VARS` on reana-workflow-controller, and
`workflow_run_manager.py` extends the sidecar environment with the parsed
values. A deployment with an egress proxy but no matching `NO_PROXY` therefore
routes this loopback-only shutdown call to the proxy. The preStop hook then
fails and the batch pod can disappear without stopping its running jobs;
Kubernetes reports only a lifecycle-hook event, making the cause difficult to
diagnose.

Suggested fix — use an opener with proxies explicitly disabled for this local
call, for example
`urllib.request.build_opener(urllib.request.ProxyHandler({})).open(...)`, and
add a test showing that injected proxy variables cannot change the shutdown
destination.

tiborsimko: **🤖 PR696-07 follow-up (2026-08-25): verified fixed at 579c4e2**

The loopback preStop command now builds an opener with an empty `ProxyHandler`
mapping before calling the job-controller shutdown endpoint. Injected
`HTTP_PROXY`/`http_proxy` settings can therefore no longer route the in-Pod
loopback request through an egress proxy. The test asserts the explicit
`ProxyHandler({})`, rejects a fallback to direct `urllib.request.urlopen()`,
and compiles the rendered command to verify its syntax. No residual concern
remains for this finding.

reana_workflow_controller/workflow_run_manager.py
```diff
@@ -767,9 +798,44 @@ def _create_job_spec(
...
+                                "    urllib.request.build_opener("
+                                "urllib.request.ProxyHandler({{}})"
+                                ").open("
+                                "'http://127.0.0.1:{}{}', timeout={}).read()\n"
+                                "except Exception:\n"
+                                "    pass"
+                            ).format(
+                                JOB_CONTROLLER_CONTAINER_PORT,
+                                JOB_CONTROLLER_SHUTDOWN_ENDPOINT,
+                                REANA_RUNTIME_BATCH_TERMINATION_GRACE_PERIOD,
```
tiborsimko: **🤖 PR696-08: The preStop request consumes the full termination grace period (LOW)**

The loopback shutdown request now uses exactly the same value as the Pod's
termination grace period:

```python
"'http://127.0.0.1:{}{}', timeout={}).read()\n"
...
REANA_RUNTIME_BATCH_TERMINATION_GRACE_PERIOD,
```

```python
termination_grace_period_seconds=REANA_RUNTIME_BATCH_TERMINATION_GRACE_PERIOD,
```

Kubernetes starts the termination-grace countdown before invoking `preStop`
and sends SIGTERM to a container only after its hook completes. If `/shutdown`
or the loopback request hangs until this socket timeout, the hook can consume
the entire budget and the job-controller process reaches the deadline without
any time to handle SIGTERM before it is forcibly killed. Kubernetes documents
that the grace period covers both the hook and normal process shutdown:
<https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/>.

Increasing the previous hard-coded ten seconds is reasonable for workflows
with several jobs, but matching the outer deadline removes the required
shutdown margin.

Suggested fix — use a dedicated request timeout below the Pod grace period,
for example `max(1, grace_period - shutdown_margin)`, and reserve an explicit
margin for hook startup, signal delivery, and process termination. Add a
manifest-generation test proving the inner timeout remains strictly below
`terminationGracePeriodSeconds` for supported configurations.

reana_workflow_controller/workflow_run_manager.py
```diff
@@ -961,10 +1038,52 @@ def _create_job_spec(
...
+                            "name": "kube-root-ca.crt",
+                            "items": [{"key": "ca.crt", "path": "ca.crt"}],
+                            # An absent ConfigMap (a non-standard cluster
+                            # setup, or a namespace where it hasn't been
+                            # provisioned yet) must not wedge every batch
+                            # pod in ContainerCreating; this projection is
+                            # a best-effort convenience, not a hard
+                            # requirement for the job-controller container
+                            # to run.
+                            "optional": True,
```
tiborsimko: **🤖 PR696-09: The optional CA projection moves failure from scheduling to runtime (NIT)**

The projected `kube-root-ca.crt` source is now optional so a missing ConfigMap
does not hold the Pod in `ContainerCreating`:

```python
"configMap": {
    "name": "kube-root-ca.crt",
    "items": [{"key": "ca.crt", "path": "ca.crt"}],
    "optional": True,
}
```

The job-controller's Kubernetes client still calls
`kubernetes.config.load_incluster_config()`. That loader requires the standard
service-account CA file and raises `ConfigException` when it is absent. The
token and downward-API projections do not supply an alternative trust bundle.
Consequently, on the non-standard cluster this change targets, the Pod can be
scheduled but the job controller fails when it first constructs its
Kubernetes client. This converts a visible volume-projection failure into a
later application failure rather than providing the documented fallback.

Suggested fix — keep the ConfigMap projection required, or implement and test
a real alternative CA/configuration path before marking it optional. Update
the comment to describe the actual dependency, and add a test for the intended
missing-ConfigMap behaviour if such clusters are meant to be supported.

tiborsimko (approved):

